PR #20, opened 2015-08-18
Fix for Sinon.JS issue #766: clearTimeout() no longer clears Immediat…

rogierschouten:
…e/Interval and vice versa

A clearTimeout(), given the ID of a timer started with setImmediate() or
setInterval() would erroneously clear that immediate or interval

### DIFF
--- a/src/lolex.js
+++ b/src/lolex.js
@@ -260,11 +260,11 @@
 
     function timerType(timer) {
         if (timer.immediate) {
-            return "immediate";
+            return "Immediate";
         } else if (typeof timer.interval !== "undefined") {
-            return "interval";
+            return "Interval";
         } else {
-            return "timeout";
+            return "Timeout";
         }
     }
 
@@ -290,7 +290,9 @@
             var timer = clock.timers[timerId];
             if (timerType(timer) === ttype) {
                 delete clock.timers[timerId];
-            }
+            } else {
+				throw new Error("Cannot clear timer: timer created with set" + ttype + "() but cleared with clear" + timerType(timer) + "()");
+			}
         }
     }
 
@@ -382,7 +384,7 @@
         };
 
         clock.clearTimeout = function clearTimeout(timerId) {
-            return clearTimer(clock, timerId, "timeout");
+            return clearTimer(clock, timerId, "Timeout");
         };
 
         clock.setInterval = function setInterval(func, timeout) {
@@ -395,7 +397,7 @@
         };
 
         clock.clearInterval = function clearInterval(timerId) {
-            return clearTimer(clock, timerId, "interval");
+            return clearTimer(clock, timerId, "Interval");
         };
 
         clock.setImmediate = function setImmediate(func) {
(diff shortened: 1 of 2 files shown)


mantoni:
This can't be merged for some reason.

rogierschouten:
Can you describe what happens? It says "This branch is up-to-date with the base branch" here and I don't know what to do

mantoni:
Funny. It's resolved. Can be merged again.

src/lolex.js
```diff
@@ -258,6 +258,42 @@
...
+
+    function clearTimer(clock, timerId, ttype) {
+        if (!timerId) {
+            // null appears to be allowed in most browsers, and appears to be
+            // relied upon by some libraries, like Bootstrap carousel
+            return;
+        }
+
+        if (!clock.timers) {
+            clock.timers = [];
```
mantoni: You could simply return here.

rogierschouten: I agree, but I did not change that code. I moved that entire function body from its old place in clock.clearTimeout() and only added the last few lines.  
That thing with setting clock.timers to [] was so specific that I thought it must have a function and I didn't want to mess with it. Want me to change it anyway?

mantoni: Ah, I see. Leave it as it is then.

mantoni:
Ok, now this is odd: I ran a couple of tests and this is how it behaves in the real world.

### Node 0.10.38

```js
var t = setTimeout(function () { console.log('!'); }, 10); clearInterval(t); // Prints !
var t = setTimeout(function () { console.log('!'); }, 10); clearImmediate(t); // Doesn't print !

var t = setImmediate(function () { console.log('!'); }); clearTimeout(t); // Prints !
var t = setImmediate(function () { console.log('!'); }); clearInterval(t); // Also prints !
```

### Chrome 43

```js
var t = setTimeout(function () { console.log('!'); }, 10); clearInterval(t); // Doesn't print !
// setImmediate and clearImmediate throw because they're not available.
```

### Firefox 31.2.0

```js
var t = setTimeout(function () { console.log('!'); }, 10); clearInterval(t); // Doesn't print !
// setImmediate and clearImmediate throw because they're not available.
```

### IE 9

```js
var t = setTimeout(function () { console.log('!'); }, 10); clearInterval(t); // Doesn't print !
// setImmediate and clearImmediate throw because they're not available.
```

That's all I have at hand at the moment. To my surprise, it turns out the current Lolex behavior is closer to reality than it would be with your fix. However, node 0.10 is even inconsistent in it's behavior. I would be very interested in how io.js behaves. I can check that later at home.

rogierschouten:
That is true, but in this case you don't want to mimick reality (I had this discussion on the sinon issue as well). It is bad programming practice to have setTimeout() followed by clearImmediate() -  and on some platforms, as you saw, it will not do what you intended. Therefore it would be handy if Sinon was strict so you can find your bug in unittests instead of discovering it on one of the platforms

mantoni:
Reasonable. Shouldn't it then even throw if you're trying to, let's say, clear an interval with `clearTimeout`? With this change it would just silently do nothing.

rogierschouten:
Good idea, done!

cjohansen:
Lolex 1.3.0 out

cjohansen:
Sinon 1.16.0 out

rogierschouten:
Awesome, thanks!!

mantoni:
:+1:

PR #22, opened 2015-08-18
Fix for Sinon.JS issue #808: add setSystemTime() function

rogierschouten:
setSystemTime() changes the current time while ensuring that the behavior of any existing timers doesn't change. It simulates the user changing the system clock while the program-under-test is running.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -104,6 +104,10 @@ var clock = lolex.install();
 
 ### `clock.tick(time)`
 
+### `clock.setSystemTime([now])`
+This simulates a user changing the system clock while your program is running.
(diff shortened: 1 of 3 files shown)


mantoni:
Nice. LGTM. @cjohansen ?

cjohansen:
Good stuff!

rogierschouten:
Thx! Will you be integrating it into the sinon package as well or do we
need to coordinate on that?

cjohansen:
I will, I thought I'd wait until you and @mantoni finish up #20

rogierschouten:
I think we're done for #20. He had a suggestion and I implemented it

PR #25, opened 2015-09-08
Add support for process.hrtime()

rogierschouten:
This won't allow nanosecond ticks yet, but at least process.hrtime() will now increase along with clock.tick().


fatso83:
In case you haven't noticed, the tests fail :-) Would you mind fixing them and try again?

fatso83:
bump.

rogierschouten:
@fatso83 sorry, I've been moving house and this has slipped my attention. Working on it now.

rogierschouten:
Unsure what to do about the lint errors from BitHound - I didn't seem to have introduced them? Can we ignore them (especially the eval() call)?

fatso83:
This looks pretty good to me, but like you, I am a bit stumped as to what we should do with the BitHound results. If the existing code already is breaking in BitHound (which it seems to do), then we should have a proper cleanup of the existing code to remove all these warnings before we refuse pull requests on the basis of what BitHound says.

@mroderick : should we try to address this?

@rogierschouten : Forgetting BitHound for a second - could you squash your changes into one commit?

.settings/settings.json
```diff
@@ -0,0 +1,5 @@
+// Place your settings in this file to overwrite default and user settings.
```
fatso83: Your IDE settings do not belong in the repo. Delete this

src/lolex-src.js
```diff
@@ -1,552 +1,606 @@
-/*global global, window*/
```
fatso83: I had to view this file with [whitespace diff turned off](https://github.com/sinonjs/lolex/pull/25/files?w=1) to view the difference. Could you configure your editor to keep the existing indenting settings? 

We already have an [EditorConfig](https://github.com/sinonjs/lolex/blob/master/.editorconfig) file. Perhaps try installing a plugin for your editor that will use this?

mroderick:
> @mroderick : should we try to address this?

Yes, I just haven't had time to finish what I started. As soon as we get `master` to green for all checks, then we can make them mandatory.

src/lolex-src.js
```diff
@@ -1,552 +1,606 @@
...
+                    timer.createdAt += difference;
+                    timer.callAt += difference;
+                }
+            }
+        };
+
+        if (hrtimePresent) {
+            clock.hrtime = function (prev) {
+                if (Array.isArray(prev)) {
+                    var oldSecs = (prev[0] + prev[1] / 1000000000);
```
mroderick: If you write large numbers using exponential notation, it will be a lot easier for readers to discover mistakes. Otherwise it's easy to overlook that `1000000000` is written as `10000000000` or `100000000`

```javascript
1000000000 == 10e8
// => true

1000000 == 10e5
// => true

1000 == 10e2
// => true
```

rogierschouten:
Thanks for your comments guys!  Next week I'll have time to process them.

rogierschouten:
whitespace - fixed
write numbers using exponential notation - done
editor settings - removed
squash commits --> created a new branch to avoid rebasing a published branch, 

see pull request #40

PR #26, opened 2015-09-10
Fix for breaking setImmediate polyfills

fatso83:
When binding the original native functions to the global object to make them writable in IE, `lolex` assigns setImmediate no matter if it is present as a native browser function or not. That made some polyfills fail, as they would add the polyfill to the _prototype of window_, not the window object itself. Thus, the property added by `lolex` would shadow that of the prototype.

The fix simply checks if the property is present beforeadding it.

Should implicitly fix issue cjohansen/Sinon.JS#821

Verify problem was introduced in 4aec26199f14df8bca46d300811cfcf9ec03ccc3
```
git clone  git@github.com:sinonjs/lolex.git 
cd lolex
curl -o test-script.sh https://raw.githubusercontent.com/fatso83/lolex/setImmediate-shim-bug/test-script.sh
git bisect start v1.2.1 v1.2.0  && git bisect run sh test-script.sh
```
produces 
> 4aec26199f14df8bca46d300811cfcf9ec03ccc3 is the first bad commit

The test [run by `test-script.sh`](https://github.com/fatso83/lolex/tree/setImmediate-shim-bug) breaks in IE9, Chrome and PhantomJS.

After the fix the test passes in all current browsers (including Chrome Canary 47.0.2503).

### DIFF
--- a/lolex.js
+++ b/lolex.js
@@ -18,11 +18,16 @@
 
     global.setTimeout = glbl.setTimeout;
     global.clearTimeout = glbl.clearTimeout;
-    global.setImmediate = glbl.setImmediate;
-    global.clearImmediate = glbl.clearImmediate;
     global.setInterval = glbl.setInterval;
     global.clearInterval = glbl.clearInterval;
     global.Date = glbl.Date;
+
+    // setImmediate is not a standard function
+    // avoid adding the prop to the window object if not present
+    if('setImmediate' in global) {
+        global.setImmediate = glbl.setImmediate;
+        global.clearImmediate = glbl.clearImmediate;
+    }
 
     // node expects setTimeout/setInterval to return a fn object w/ .ref()/.unref()
     // browsers, a number.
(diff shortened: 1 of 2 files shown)


zowers:
> they would add the polyfill to the prototype of window, not the window object itself
not always, YuzuJS/setImmediate polifyll checks where setTimeout lives and adds setImmediate there
https://github.com/YuzuJS/setImmediate/blob/master/setImmediate.js#L147 :
```
    // If supported, we should attach to the prototype of global, since that is where setTimeout et al. live.
    var attachTo = Object.getPrototypeOf && Object.getPrototypeOf(global);
    attachTo = attachTo && attachTo.setTimeout ? attachTo : global;
```
so it is `window.__proto__` in Chrome and IE but `window` in Firefox and Edge

fatso83:
@zowers do observe the part of the sentence: made _some_ polyfills fail

This PR fixes the problems seen in Chrome and PhantomJS with SinonJS and YuzuJS' setImmediate polyfill.

zowers:
having a test (and fix) which works only in some environments is even worse than having exception in all environments.
"Chrome is new Internet Explorer" future is getting closer with steps like this.

fatso83:
I am not sure what you are getting at? The fix in itself is not targetting any specific browsers, it just happens to fix the problem in the browsers that are affected. `lolex` had a bit of code that was made in way that made the built-in functions overridable in every browser. The only problem with that code was that it did not use a defensive approach: there was no check if the browser actually had the function it was trying to make settable. If you believe that the problem is of a more global nature and that one should use some kind of `makeOverridableIfDefined('setTimeout)` every time, then just say so(, although I think that is kind of pointless here as it's just one property that is non-standard). Guessing games are boring.

So, instead of sniping, I propose you just submit a PR to my branch with whatever changes you would like to see. As long as I am not sure what you are objecting to having this discussion is kind of pointless and fills up the inbox of others subscribers.

zowers:
sorry for being harsh, I just wanted to point that the polyfill and browsers all have different variants of how they define the functions, so it's better to verify that the fix is valid in all those variants.

fatso83:
> verify that the fix is valid in all those variants.

That statement is a bit ambiguous, so I am not quite sure what you mean. Do you mean simply testing if `lolex` works as advertised in all current browsers after the fix? If so; totally agree, and I have already done that. The test script mentioned above passes in Firefox, Chrome, Internet Explorer 9, Opera and Phantom. Before the fix only Firefox passes.

If you on the other hand mean customizing a fix for each browser and polyfill, I find it bordering on absurd. The fix is merely a guard clause around a destructive operation: it is not targeting any specific browser. All all browsers will use the implementation at the lowest point in the prototype chain, so it does not matter where the actual implementation (if one exists) is actually defined. The point of `lolex` is to give you fake timers, period. Implementation details of any polyfills is irrelevant: just don't add timers that are not already there.

zowers:
> lolex works as advertised in all current browsers 
yes, exactly that

>The test script mentioned above passes in Firefox, Chrome, Internet Explorer 9, Opera and Phantom. great, thanks!

mroderick:
Very nice work @fatso83. Would you mind squashing it down to just one commit?

cjohansen:
Nice work! :+1:

fatso83:
Squashed  :)

PR #28, opened 2015-11-27
Fix lint errors

novemberborn:
`npm run lint` was failing. These seem to be the minimum changes required for it to be happy. I've tried fixing lint errors in the test file as well but there are two cases however that jslint won't ever be happy about. I've added a comment explaining to comment those out.

Perhaps a different linter should be used?

I've added linting (of the source file) to the CI script to ensure the source stays nice and tidy.

### DIFF
--- a/script/ci-test.sh
+++ b/script/ci-test.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -eu
 
+npm run lint
 npm run test-node
 npm run test-headless
--- a/src/lolex.js
+++ b/src/lolex.js
@@ -22,7 +22,7 @@
 
     // setImmediate is not a standard function
     // avoid adding the prop to the window object if not present
-    if('setImmediate' in global) {
+    if (global.setImmediate !== undefined) {
         global.setImmediate = glbl.setImmediate;
         global.clearImmediate = glbl.clearImmediate;
     }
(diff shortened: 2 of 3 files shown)


mroderick:
Thanks!

I agree, we should switch to using ESLInt in all repos under sinon organisation

novemberborn:
> I agree, we should switch to using ESLInt in all repos under sinon organisation

Is <https://github.com/sinonjs/sinon/blob/3fb8e6ad75e0e0f06e8aea0941917c889693d698/.eslintrc> the canonical version? Maybe I can give that a go.

mroderick:
> Is https://github.com/sinonjs/sinon/blob/3fb8e6ad75e0e0f06e8aea0941917c889693d698/.eslintrc the canonical version? Maybe I can give that a go.

TBH, I would rather like to see us create a fork of https://github.com/walmartlabs/eslint-config-defaults, with a Sinon configuration. I've created an https://github.com/sinonjs/sinon/issues/946 for that. If you want to take a stab at it, just let me know, and I'll import your repository.

novemberborn:
> If you want to take a stab at it, just let me know, and I'll import your repository.

I'd rather not actually, sorry. Pretty happy with `npm install --save-dev standard` myself :wink:

PR #29, opened 2015-11-27
Update readme and some code description fixes

novemberborn:
* `requestAnimationFrame` was mentioned but I don't see any code support for it. I've removed it from the docs, but please shout if I overlooked it.
* Updated the "Faking the native timers" section so it leads with the default behavior and better explains "context".
* Updated the "Running tests" section
* Added far more complete API docs, including allowed values for the various arguments.
* I've changed `npm test` to not run `test-cloud` since that requires Sauce Lab access.
* Fixed the `parseTime()` error message which mentioned fewer options than were allowed.
* Corrected some time related test descriptions where the description didn't match the time in the test itself.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -125,7 +125,7 @@ It affects the current time but it does not in itself cause e.g. timers to fire;
 ## Running tests
 
 Lolex has a comprehensive test suite. If you're thinking of contributing bug
-fixes or suggest new features, you need to make sure you have not broken any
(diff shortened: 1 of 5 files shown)


mroderick:
Looks good! Of course, since I merged the lint branch, this now needs a rebase.

novemberborn:
@mroderick rebased.

PR #42, opened 2016-02-28
Run test-cloud on Travis

mantoni:
We already have the possibility to run tests in real browsers with `npm run test-cloud`. The prerequisite to be able to run this command is a free SauceLabs account and the `SAUCE_USERNAME` and `SAUCE_ACCESS_KEY` environment variables.

If these environment variables are configured in the Travis config, SauceLabs can be used from there. With this config, we run the `test-cloud` build only once in the node 4.2 build.

I used my SauceLabs account for the purpose of trying this out. We can keep it this way or create a „Sinon“ account which we can share. The access key was encrypted using the method described [in the Travis documentation](https://docs.travis-ci.com/user/encryption-keys/).

The first successful build (from this branch) is here: https://travis-ci.org/sinonjs/lolex/jobs/112401049

For some unknown reason the Chrome build returns a status code 102 every single time. I have this running successfully in another project, but it doesn’t work reliably with Lolex. However, the build passes on latest Firefox and IE 9, 10 and 11.

We can add more browser configs of cause in the `.min-wd` file.

Any thoughts?

### DIFF
--- a/.min-wd
+++ b/.min-wd
@@ -1,7 +1,7 @@
 {
   "sauceLabs": true,
   "browsers": [{
-    "name": "chrome"
+    "name": "firefox"
   }, {
     "name": "internet explorer",
     "version": "9",
@@ -9,5 +9,8 @@
   }, {
     "name": "internet explorer",
     "version": "10"
+  }, {
+    "name": "internet explorer",
+    "version": "11"
   }]
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,13 @@ node_js:
   - "0.12"
   - "4.2"
   - "5"
+
+env:
+  global:
+    - SAUCE_USERNAME=sinonjs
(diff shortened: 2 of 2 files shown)


fatso83:
Good now is better than perfect never. We'll add Chrome later. Good work!

mantoni:
@cjohansen Can you set up an email address `@sinonjs.org` and use it to create a SauceLabs account for Sinon? Then we can share the account information and I can update this PR with the corresponding `SAUCE_ACCESS_KEY`.

fatso83:
@mantoni: the builds have been failing for a while. I think this PR never worked on Travis.

The [current failing build](https://travis-ci.org/sinonjs/sinon/jobs/117175966) says
>     You used username 'sinonjs' and access key 'None' to authenticate, which are not valid Sauce Labs credentials.

I looked into the [Travis documentation](https://docs.travis-ci.com/user/sauce-connect/#Setting-up-Sauce-Connect) and I think we are missing a line. What should have been:

```
   access_key:
      secure: "The secure string output by `travis encrypt`"
```

is just 

```
      secure: "The secure string output by `travis encrypt`"
```

in our travis.yml.

fatso83:
That comment was supposed to go in `sinon` issues, not `lolex` (where builds are working). I blame Safari :stuck_out_tongue: 

I'll make a separate issue on sinon. Regarding the difference in documentation, there is more than one way of configuring SauceLabs, I found ... I was looking at [Sauce Connect](https://docs.travis-ci.com/user/sauce-connect) on the Travis page, while you probably read [this bit about encrypting env variables](https://docs-staging.saucelabs.com/ci-integrations/travis-ci/#adding-credentials-for-a-public-github-repo) on Sauce Labs :smile:

PR #43, opened 2016-03-07
Add tests about Date creation

fraisse:
Add tests demonstrating that a fake Date could be created with one argument as a String since this string is in a format recognized by the Date.parse() method.


fatso83:
Thanks! Tests are documentation too :-)

PR #45, opened 2016-03-23
Add runAll method to run timers until empty

karl:
This runs all pending timers until there are none remaining. If new timers are added while it is executing they will be run as well.

This makes it easier to run asynchronous tests to completion without worrying about the number of timers they use, or the delays in those timers.

It runs a maximum of `loopLimit` times after which it assumes there is an infinite loop of timers and throws an error.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -97,15 +97,18 @@ without arguments.
 
 ## API Reference
 
-### `var clock = lolex.createClock([now])`
+### `var clock = lolex.createClock([now[, loopLimit]])`
 
 Creates a clock. The default
-[epoch](https://en.wikipedia.org/wiki/Epoch_%28reference_date%29) is `0`. Now
-may be a number (in milliseconds) or a Date object.
+[epoch](https://en.wikipedia.org/wiki/Epoch_%28reference_date%29) is `0`.
 
-### `var clock = lolex.install([context[, now[, toFake]]])`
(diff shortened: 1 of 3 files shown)


fatso83:
Seems like a useful addition. Not super sure about the telescoping method signatures, but I think it's alright for now. Merging.

fatso83:
Ah, damn. I cannot merge this as long as the status checks are failing. 

@mantoni : did we not conclude that running the tests on PRs would fail in the SauceLabs config? If so, could you somehow disable SauceLabs for PRs?

mantoni:
@fatso83 Yes, [I did this for Sinon](https://github.com/sinonjs/sinon/commit/66123fea53b8fe43cb3acba9ded7c3a115c8ba2a) but not yet for Lolex. You can quickly do the same change here, or I can do it later tonight.

fatso83:
@mantoni Thanks for the link. Made PR #46 now to address it. Will probably just merge it myself once the checks complete.

mantoni:
@karl Can you please rebase to pull in the fix to get the build passing?

karl:
@mantoni Done!

karl:
@mantoni, @fatso83: I just realised that previously I'd just merged in `master`. I've correctly rebased the PR on to `master` now, so should be ready to merge at your convenience!

mantoni:
Cool, thanks!

Issue #44, opened 2016-03-22
Add methods to run all timers and run all currently pending timers


karl:
Facebook's Jest testing library has a couple of methods that make it easy to run all the timers:

## Run All Timers

* https://facebook.github.io/jest/docs/timer-mocks.html#run-all-timers

This runs all pending timers until there are none remaining. If new timers are added while it is executing they will be run as well.

## Run Pending Timers

* https://facebook.github.io/jest/docs/timer-mocks.html#run-pending-timers

This runs all the pending timers until there are none remaining. Any new timers that are added while it is executing are *not* run.

---

Would there be any interest in adding similar functionality to lolex? I'm happy to look into doing the work if so.

novemberborn:
There is `.next()` which runs the next timer. You could call it repeatedly until the clock stops advancing. This would cover the "run all timers" case.

I suspect for the pending timers you may be able to uninstall the current clock and install a new one, then use `.next()` on the old clock.

karl:
I've created a PR with an implementation of a `runAll` method.

If there is interest in merging it and there are any changes/tweaks you'd like me to make just let me know. I've attempted to follow the lolex code style where I can.

karl:
The `runAll` method has now been merged in to master :smile:

karl:
I've created a second PR with a `runToLast` method.

This takes note of the last scheduled timer when it is run, and advances the
clock to that time firing callbacks as necessary.

If new timers are added while it is executing they will be run only if they
would occur before this time.

This is useful when you want to run a test to completion, but the test recursively
sets timers that would cause `runAll` to trigger an infinite loop warning.

I decided to take this approach, rather than Jest's `runOnlyPendingTimers` approach to avoid having a situation where timers could be called out of order.

fatso83:
Both your PRs are now merged. This should conclude this issue, right? Close?

karl:
Yep, thanks for being open to the changes!

PR #50, opened 2016-04-06
Run to last existing timer

karl:
Add the ability to run to the last currently scheduled timer.

https://github.com/sinonjs/lolex/issues/44

This is useful when you want to run a test to completion, but the test recursively sets timers that would cause `runAll` to trigger an infinite loop warning.


mantoni:
Nice! LGTM.

PR #53, opened 2016-04-08
Check for existence of `process` before using it

karl:
This fixes a bug where Lolex would no longer work in browsers because `process` doesn't exist.

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -405,8 +405,11 @@
         setInterval: setInterval,
         clearInterval: clearInterval,
         Date: Date,
-        hrtime: global.process.hrtime
     };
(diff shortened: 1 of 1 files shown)


fatso83:
LGTM, but I don't think we want the built code as well. Or do we, @mantoni? In my view it should only be updated once we release a new version.

mantoni:
@fatso83 Agreed. The "compiled" output should only update for releases.

fatso83:
@karl could you remove the compiled output and squash the resulting commits?

karl:
@fatso83, @mantoni: I've removed the compiled output, so hopefully all good to go!

Issue #59, opened 2016-05-25
Cannot set property 'hrtime' of undefined


gautaz:
Hello,

After migrating from lolex 1.4.0 to 1.5.0, my tests fails with this exception:
```javascript
     TypeError: Cannot set property 'hrtime' of undefined
      at uninstall (node_modules/lolex/src/lolex-src.js:360:39)
      at Object.clock.uninstall (node_modules/lolex/src/lolex-src.js:629:13)
      at bus.close.then.then (test/scheduling/testScheduler.js:64:57)
```

I'm using nodejs 6.2.0.

My tests use lolex this way:

Setup:
```javascript
clock = lolex.install(context)
```

Tear down:
```javascript
clock.uninstall()
```

[lolex-src.js 1.5.0](https://github.com/sinonjs/lolex/blob/v1.5.0/src/lolex-src.js#L360) seems to assume that the context object has a key named `process`.

Any idea regarding this ? Does this mean that compatibility was broken somehow ?

fatso83:
@rogierschouten Could this be related to #40 ?

gautaz:
You are right, the breaking code was added by this merge.

Doesn't this means that 1.5.0 breaks backward compatibility and should be tagged as 2.0.0 (a migration note can also probably help)?
Or perhaps the code should be modified to not rely on target.process?

fatso83:
@gautaz This breaking change was unintentional. Needs a bugfix.

rogierschouten:
looking into it now

rogierschouten:
Issued PR #62 for this. I could not reproduce the issue in any way also not in a browser. However, the code was clearly not airtight so I think this would fix it.

fatso83:
@gautaz Could you please provide a test case that is runnable in Node? I could then squash that together with #62 if that indeed fixes it.

gautaz:
Here it is: #65

fatso83:
@rogierschouten Mind seeing if @gautaz test breaks for you? If so, then you have your reproducible test case. You could then merge that into #62 (just remove the built source).

rogierschouten:
@fatso83 done

fatso83:
Fixed. Will release a new version asap.

PR #62, opened 2016-05-26
fix for issue #59

rogierschouten:

PR #65, opened 2016-05-27
test case for #59

gautaz:
Bear in mind that this is the way I was using lolex until revision 1.4.0, the additional test case fails for lolex 1.5.0 with nodejs 6.2.0.
Do not hesitate to comment if I am misusing lolex from the beginning.

### DIFF
--- a/test/lolex-test.js
+++ b/test/lolex-test.js
@@ -29,6 +29,23 @@ var GlobalDate = Date;
 var NOOP = function NOOP() { return undefined; };
 var hrtimePresent = (global.process && typeof global.process.hrtime === "function");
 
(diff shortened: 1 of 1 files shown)


fearphage:
What is the point of these tests? What are you trying to confirm? You're not making any assertions. You need to confirm something does/doesn't exist, has/hasn't changed, or did/didn't throw.

rogierschouten:
I've incorporated this test (in slightly different form) in PR #62 which fixes issue #59. This test reproduces #59.

fatso83:
Merged in #74 (author info changed to reflect your contribution).

PR #68, opened 2016-06-10
Correct the clear{Interval|Timeout|Immediate} error message when attempting to clear a different type of timer.

Standard8:
The error message given when a timer created by `setTimeout` is cleared by a different type is wrong - the `set* / clear*` types are the wrong way around.

Additionally the test cases for the exceptions are currently not testing the code correctly - currently the error thrown is `Cannot read property 'clock' of undefined` due to the missing bind.

This fixes the error message and tests that its correct for all the combinations.


mantoni:
Thank you!

PR #70, opened 2016-06-14
fixed a case where runAll was called and there are no timers

CurtisHumphrey:


fatso83:
Had to run it without your fixes to see that this really wasn't fixed already, and lo and behold: it didn't. So thanks!

CurtisHumphrey:
you are welcome! I happen to have a case where only sometimes a timer was set which is how I caught it.

PR #75, opened 2016-07-13
Replace JSLint with ESLint

fatso83:
I got tired of all the BitHound linting warnings due to it not supporting JSLint, so instead of disabling those checks, I decided it would be better to just replace JSLint with its successor. The rules are copied from Sinon. Some needed to go, while a few were modified. 

Most changes deal with whitespace and quotation marks (fixed by `eslint --fix`).

There are now just a few warnings left (on the use of `++` and `--`). 
<hr />
Unfortunately, I found out first after the PR was submitted that BitHound just supports ESLint rules for version 1? Therefore I had to disable linting for BitHound. No biggie as we do it anyhow.


.bithoundrc
```diff
@@ -1,54 +1,15 @@
...
-    "**/extjs/examples/**",
-    "**/extjs/locale/**",
-    "**/extjs/packages/**",
-    "**/extjs/plugins/**",
-    "**/extjs/resources/**",
-    "**/extjs/src/**",
-    "**/extjs/welcome/**",
-    "bower_components/**"
+    "script",
+    "**/node_modules/**"
```
mantoni: I think `node_modules` is ignored by default. This shouldn't be necessary,

mantoni:
:+1: Looking good.

PR #77, opened 2016-07-19
Fix setInterval() behavior with string times

fdnk:
In Node.js, when setTimeout is called with an time string, it is parsed. If the string is an invalid number, it's assumed as 0. In lolex, using a string as time causes the setTimeout not to trigger the callback function.

This is the behaviour that I would expect:

```js
it("parses no-numeric string times", function () {
    this.clock.setTimeout(function () { lolex.evalCalled = true; }, "string");
    this.clock.tick(10);

    assert(lolex.evalCalled);
});

it("evals non-function callbacks", function () {
    this.clock.setTimeout("lolex.evalCalled = true", 10);
    this.clock.tick(10);

    assert(lolex.evalCalled);
});
```

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -173,7 +173,7 @@
 
         timer.id = uniqueTimerId++;
         timer.createdAt = clock.now;
-        timer.callAt = clock.now + (timer.delay || (clock.duringTick ? 1 : 0));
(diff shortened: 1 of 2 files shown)


fatso83:
Thanks!

PR #81, opened 2016-08-08
Update installation notes

fatso83:
Updates some installation notes. Removes the reference to it being hosted on the Sinon site. Has not been true for quite a while. Also adds some info on using it in a browser.


mantoni:
Looks good. Just that npm is officially written "npm", not "NPM" because they say it's not an acronym.

fatso83:
First edit, commit and squash on my phone. Not pleasant, but not too shabby either :-)

PR #86, opened 2016-11-05
Pass eslint warnings

duncanbeevers:
Use `+= 1` in favor of `++`
Disable no-plusplus rule for `while (i--)` loop, rather than try and assign within a comparison
Allow deeply-nested callbacks in test file


cjohansen:
Thanks for the effort, but we're now changing the eslint config to match the one in sinon... So these don't apply anymore.

duncanbeevers:
No prob

PR #89, opened 2016-12-15
Allow install to be called with date object

### LINKED ISSUE #73: Installing a clock does not work with a Date object
As mentioned in the docs: https://github.com/sinonjs/lolex#var-clock--lolexcreateclocknow-looplimit

> The now argument may be a number (in milliseconds) or a Date object.

However, this does not work for a Date object.

It seems as though this line (https://github.com/sinonjs/lolex/blob/f464603ef78b585aabb1ad461b26a3267989536c/src/lolex-src.js#L620) needs to be modified to:

``` js
if (typeof target === 'number' || target instanceof Date) {
  // ...
}
```

th3fallen:
Fixes #73

### DIFF
--- a/lolex.js
+++ b/lolex.js
@@ -616,7 +616,7 @@
         var i,
             l;
 
-        if (typeof target === "number") {
+        if (typeof target === "number" || target instanceof Date) {
             toFake = now;
             now = target;
             target = null;
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -614,7 +614,7 @@
         var i,
             l;
 
-        if (typeof target === "number") {
+        if (typeof target === "number" || target instanceof Date) {
             toFake = now;
             now = target;
             target = null;
--- a/test/lolex-test.js
+++ b/test/lolex-test.js
@@ -45,7 +45,8 @@ describe('issue #73', function() {
     it('should install with date object', function () {
         var date = new Date('2015-09-25');
         var clock = lolex.install(date);
-        assert.same(date.clock, clock);
+        assert.same(clock.now, 1443139200000);
+        clock.uninstall();
     });
 });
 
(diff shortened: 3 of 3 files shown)


th3fallen:
Not sure about the test however :(

fatso83:
Five breaking tests with half a line of diff, that is impressive :-D Seriously, you might look into setting a breakpoint in some of those tests to inspect what is going on. I cannot see it from just glancing, at least.

th3fallen:
what can i say im good at breaking things. i'll take a look at it tomorrow morning.

th3fallen:
@fatso83 so I've got a question, I've fixed the tests but I notice this test https://github.com/sinonjs/lolex/blob/78323dc3ff6b69c76dcc9a9b43a106c9859d3d7a/test/lolex-test.js#L1720-L1725 that tests the exact opposite of what I'm trying to achieve here? is that intended? or can i just drop that test?

fatso83:
Sorry about not answering that for so long, btw. Not sure, I'll try to `git blame` it to get some context this week (holidays).

fatso83:
totally forgot about this, and I have been way too busy (and is). hoping someone can look into this...

fatso83:
OK, I am terrible at letting things go, so I went down the rabbit hole of `git blame`, back to the time `lolex` was first extracted from Sinon, and then further back to when this was first introduced as part of a fix to sinon/sinonjs#49 . Please check out [the conclusion on the discussion there](https://github.com/sinonjs/sinon/issues/49#issuecomment-3281006).

If I understand the discussion correctly, it seems as if @cjohansen made the call that since the original `Date()` constructor is not supposed to accept a `Date` object, then neither should we. Not sure I am totally following that logic, as `sinon.useFakeTimers` could well accept a date and then extract the milliseconds out of that object. Not that you are doing that here, anyway, so ... 

This requires a deeper look into the Lolex code to see how things are interconnected so we don't introduce a bug by allowing Date objects where there shouldn't be one.

cjohansen:
This has been manually applied to master. Not sure why Github doesn't pick it up.

PR #92, opened 2017-01-13
Fixed typo in clock.runAll error

### LINKED ISSUE #93: Typo in clock.runAll
Error would previously say `"Aborting after running 1000timers, assuming an infinite loop!"` (notice the lack of space between `1000` and `timers`). There should be a space there.

JoshuaKGoldberg:
Error would previously say `"Aborting after running 1000timers, assuming an infinite loop!"` (notice the lack of space between `1000` and `timers`). Adding a space there.

Fixes #93.

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -551,7 +551,7 @@
                 clock.next();
             }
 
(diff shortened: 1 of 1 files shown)


fatso83:
thanks!

Issue #85, opened 2016-10-21
Time travel without freezing


jvdp:
I'd like to use this library for acceptance tests (ie. running the web app + JS frontend in a browser / phantomjs) that involve time-dependent data, so the clock needs to be set back to a certain date. My app also uses lodash' `debounce` function, however, which relies on the passage of time. Because I don't really have a good place to put a `tick` statement the clock shouldn't be 'frozen' at a particular moment but it should advance as normal from the system date specified.

This would like the `travel` method of [Timecop.js](https://github.com/jamesarosen/Timecop.js) (which is [deprecated](https://github.com/jamesarosen/Timecop.js/issues/28#issuecomment-227208629) in favor of sinon / lolex.)

fatso83:
I don't see how this could be implemented without async behaviour, which means you just might as well just put `tick` calls in some endless `setInterval` or `process.nextTick` loop yourself.

jvdp:
We can take a timestamp of when time travel starts and then compute the difference to determine the adjust "current" time. Timecop.js does this: https://github.com/jamesarosen/Timecop.js/blob/master/lib/TimeStackItem.js

fatso83:
True, just computing the `now`value based on the diff is easily done. Such a feature should be quite easily implemented. Must have been to late to think clearly 😄  Care to take a stab at it?

jvdp:
Will see what I can do.

mroderick:
Has this issue stalled?

fatso83:
Will be closed by #102

fatso83:
closed by #102

PR #102, opened 2017-06-28
lolex can now attach itself to the system timers and automatically ad…

elad-nach:
in regards to issue #85: these changes will allow lolex to attach itself to the system timers when the clock is being installed.

when installing - lolex will set up a normal 'setInterval' on the real system timer (the one that's being stored at _setInterval) at a default value of 20ms, since every time the interval gets triggered the event loop will get blocked - I thought that 50hz is about enough to get good resolution but also not to clog the event loop at the same time (0.5-1 second is too much IMHO, 1ms too fast). every time the setInterval gets triggered - it will tick(20) thus resulting in timers getting invoked.

when uninstalling - lolex will remove the setInterval that was created.

i implemented the possibility to toggle this feature on and off through **exports.shouldAdvanceTime** - since I'm not sure that adding another argument to the constructor would be a wise choice for this use case - too much magic reordering arguments at the ctor and I would like to propagate this into sinon (which uses **install(now || 0, methods)**).

### DIFF
--- a/History.md
+++ b/History.md
@@ -6,6 +6,8 @@ v2.0.0 / 2017-07-13
   * Add support for performance.now (#106)
   * Fix issue with tick(): setSystemClock then throw
   * Update old dependencies
+  * Added support to automatically increment time (#85)
+  * Changed internal uninstall method signature
 
 v1.6.0 / 2017-02-25
 ===================
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -35,7 +35,6 @@ clearTimeout(timeoutResult);
 var NativeDate = Date;
 var uniqueTimerId = 1;
 
-var TIME_TICK = 20; // increment the mocked timer on this real interval (ms)
 /**
  * Parse strings like "01:10:00" (meaning 1 hour, 10 minutes, 0 seconds) into
  * number of milliseconds. This is used to support human-readable strings passed
@@ -349,7 +348,7 @@ function clearTimer(clock, timerId, ttype) {
     }
 }
 
-function uninstall(clock, target) {
+function uninstall(clock, target, config) {
     var method,
         i,
         l;
@@ -362,7 +361,7 @@ function uninstall(clock, target) {
         } else {
             if (target[method] && target[method].hadOwnProperty) {
                 target[method] = clock["_" + method];
-                if (method === "clearInterval" && exports.shouldAdvanceTime === true) {
+                if (method === "clearInterval" && config.shouldAdvanceTime === true) {
                     target[method](clock.attachedInterval);
                 }
             } else {
@@ -403,8 +402,8 @@ function hijackMethod(target, method, clock) {
 
 exports.shouldAdvanceTime = false;
 
-function doIntervalTick(clock) {
-    clock.tick(TIME_TICK);
+function doIntervalTick(clock, advanceTimeDelta) {
+    clock.tick(advanceTimeDelta);
 }
 
 var timers = {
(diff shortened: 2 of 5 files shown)


fatso83:
Thank you for this contribution! At first I thought this is really two pull requests in one I see, as it both implements async set* methods and real time advance. They are not necessarily the same, so I thought it might make sense to keep the config for the two separate. 

For instance, you can have the faked `Date.now()` delegate to the real function and give the elapsed time since the start without any asynchronicity. 

Then I realized that this meant the `tick` method wouldn't make sense in that scenario, and implicitly neither would timers (as you could never leave the main execution thread), so it would really only make sense together with the async behavior you added ... :blush: 

I'll leave this here for @mroderick to see as well, but :+1: from me.

src/lolex-src.js
```diff
@@ -394,6 +398,12 @@ function hijackMethod(target, method, clock) {
     target[method].clock = clock;
 }
 
+exports.shouldAdvanceTime = false;
```
fatso83: not super in-love with this, but on the other hand I also hate magic constructor argument matching. What's wrong with config objects?

elad-nach: I'm also not super in-love with this :rofl: 
is this what you mean?

```
exports.config = {
    shouldAdvanceTime: false
}
```
and then toggle it through:
```
lolex.config = {shouldAdvanceTime: true};
var clock = lolex.install(date);
```

I didn't want to overload the constructor with another parameter since I didn't dive too deep into what would be the implications of rewriting the logic doing the parameter swapping magic and what would be the possible repercussions for people using this library on contexts different than sinon. Also if adding another parameter to the install function (obviously at the end of the param list - `install(target, now, toFake, loopLimit, shouldAdvanceTime)`) - now install would be invoked with `install(date,null,null,null,true)`? not sure if it makes much sense either, and not really readable at first/second/tenth sight

fatso83: I think a function with more than a couple of optional arguments is broken anyhow, especially when the optional ones can precede others like in `install`. I wonder if it would make sense to change the Lolex API for a version 2 now, before adding yet another config option to something like `lolex.install(config)`. That would scale with new options, such as this.

mroderick:
@elad-nach would you mind providing a usage example? It would greatly help me wrap my head around how this is expected to work

elad-nach:
@mroderick sure.

The thing is that today with the current usage of lolex, if a test depends on a timer executing - there's no way to know when to tick() the timers. if you set a timeout after the clock has been installed - it won't get triggered. if you set a loop to tick() periodically - the timer won't be set because the execution context is inside the for loop, and the code that should be setting the timer does not get executed as of such.

Y need to tick() just after the timer has been set. this is not possible without either stubbing or proxying methods, since you have to use an interrupt style method to pull the program counter out of the main execution context and into a context which ticks() the lolex timer, invoking any pending timers, then returns to execute the code being tested.
issue #85 will be addressed with this fix. [this one too](https://github.com/expressjs/express/issues/2447)

src/lolex-src.js
elad-nach: I agree. well that's up to you. I'm not sure if you have any releases planned but we can definitely refactor that part (and possibly others too).
Would be happy to pitch in for this too. Maybe you could point out other parts that need patching too.

fatso83: This has now been done. Care to move the config?

fatso83:
I would really like to merge this, but other than the pretty minor config details mentioned above, the only real thing missing is documentation on how this works for new users.

@kylefleming has already done [a lot of the work](https://github.com/sinonjs/lolex/pull/105#issuecomment-314511849) on explaining the use-cases and differences between his async stuff and this, so that might help out.

Thanks a lot for the work you put into this!

elad-nach:
@fatso83 I'm on it 👍

fatso83:
Hi, Elad. I just noticed the docs were missing, so made #109. Whoever makes the docs is irrelevant to me - I am happy it just gets done by someone :)

fatso83:
P.S. If you could rebase your branch to get rid of the merge commits when you are finished working on it it would be :+1:

benjamingr:
Hey, looking at the diff I think this landed, right?

fatso83:
@benjamingr not sure what you mean. It's not merged. But the current diff is messed up so don't assume anything based on that. Not sure what has happened, but nothing an interactive rebase won't fix 😋

benjamingr:
@fatso83 oh, cool - I got confused by the commit diff. Sorry!

elad-nach:
@fatso83, i updated the readme and history. please review the changes.

also, readme line 80 - to hijack timers in another context - lolex-src line 643 - as I understand this translates into the target object in the config. please verify. 

if you want me to clean the commits I could just patch a new fork and close this PR, as the initial commits were made before the new branches were merged (or just squash them when you merge the PR).

History.md
```diff
@@ -6,6 +6,8 @@ v2.0.0 / 2017-07-13
   * Add support for performance.now (#106)
   * Fix issue with tick(): setSystemClock then throw
   * Update old dependencies
+  * Added support to automatically increment time (#85)
```
fatso83: that's very considerate, but you can drop this. we create this semi-automatically when crafting a new release using `git-extras`

PR #103, opened 2017-06-29
Change lolex.install signature

fatso83:
As shown in #102, something as simple as adding one additional configuration option becomes unmanageable due to the current design of a "telescoping method signature" where all the arguments are optional. This makes contributors reach for alternate ways of configuring lolex, such as adding flags on the object.

I propose we change the signature to accept a single configuration object with named parameters to simplify this. This PR is that proposal implemented. There are also [many other pros to constructor-like functions](https://gcanti.github.io/2014/09/25/six-reasons-to-define-constructors-with-only-one-argument.html) like this of 1-arity.

This is a breaking change and will require a new major version.

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -82,6 +82,7 @@ function fixedModulo(n, m) {
 
 /**
  * Used to grok the `now` parameter to createClock.
+ * @param epoch {Date|number} the system time
  */
 function getEpoch(epoch) {
     if (!epoch) { return 0; }
@@ -423,6 +424,10 @@ var keys = Object.keys || function (obj) {
 
 exports.timers = timers;
 
+/**
+ * @param now {Date|number} the system time
+ * @param loopLimit {number}  maximum number of timers that will be run when calling runAll()
+ */
 function createClock(now, loopLimit) {
     loopLimit = loopLimit || 1000;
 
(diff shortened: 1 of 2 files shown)


elad-nach:
I think that leaving the possibility for 
```lolex.install("2015-01-01")```
OR
```lolex.install(epochValue OR Date object)```
should still be maintained. the changes would and should break anything above 1 parameter to the install function but maybe the support for different default initialisations should be kept?

**Comment by fatso83** [comment]
I see your point, but I dunno ... IMHO if we break the signature, we might as well make it consistently break, not just surprisingly break for some of the combinations.

**Comment by fatso83** [comment]
@mroderick do you have any input on this? there was a sudden influx of changes, and so I thought this might be a good time to add this before accepting some of them.

File: src/lolex-src.js [hunk 1]
```diff
@@ -611,32 +616,25 @@ function createClock(now, loopLimit) {
...
-        target = null;
-    }
-
-    if (!target) {
-        target = global;
-    }
+/**
+ * @param config {Object} optional config
+ * @param config.target {Object} the target to install timers in (default `window`)
+ * @param config.now {number|Date}  a number (in milliseconds) or a Date object (default epoch)
```
mroderick: Is this parameter optional? Should it have a default value?

src/lolex-src.js
```diff
@@ -611,32 +616,25 @@ function createClock(now, loopLimit) {
...
+ * @param config.toFake {string[]} names of the methods that should be faked.
+ * @param config.loopLimit {number} the maximum number of timers that will be run when calling runAll()
+ */
+exports.install = function install(config) {
+    config = typeof config !== "undefined" ? config : {};
 
-    var clock = createClock(now, loopLimit);
+    var i, l;
+    var target = config.target || global;
+    var clock = createClock(config.now, config.loopLimit);
```
mroderick: What happens if `config.now` is undefined?

src/lolex-src.js
fatso83: All the current parameters are optional, `lolex.install()` is used in the README, so this behaviour-preserving and should not break anything.

We could add more explicit defaults of course, before sending them on to `createClock` and such (that have their own defaults), but I felt that would perhaps obfuscate my intent in this PR. Could add that in another commit, of course.

src/lolex-src.js
fatso83: The same [as today](https://github.com/fatso83/lolex/blob/10c053fa919141297ef77703468692b7ae6a7fc3/src/lolex-src.js#L88). It is converted to 0 down the line. 

I am not a super-fan of the current digging-down-into-the-code to see if not passing a number is allowed or not, so adding defaults higher up along with strict checking further down, is more my style.

src/lolex-src.js
mroderick: Faking time is confusing enough as it is, let's not change the api too much :)

mroderick:
:shipit:

fatso83:
Not totally sure what the squirrel with hat means ... Is that an approval? 😂

mroderick:
https://www.quora.com/GitHub-What-is-the-significance-of-the-Ship-It-squirrel

mroderick:
I am happy with this PR, merge it!

PR #104, opened 2017-07-06
Fix issue with tick(): setSystemClock then throw

kylefleming:
There is an issue with `tick()` when `setSystemClock` is called in a timer, and then that same timer throws an error. The correction that `tick()` does to account for `setSystemClock` happens within the try catch block, and not after it. If an error is thrown then this check is skipped.

The included unit tests demonstrate the issue.

### DIFF
--- a/test/lolex-test.js
+++ b/test/lolex-test.js
@@ -637,6 +637,69 @@ describe("lolex", function () {
             var value = clock.tick(200);
             assert.equals(clock.now, value);
         });
+
+        it("is not influenced by forward system clock changes", function () {
+            var clock = this.clock;
+            var callback = function () {
+                clock.setSystemTime((new clock.Date()).getTime() + 1000);
+            };
+            var stub = sinon.stub();
+            clock.setTimeout(callback, 1000);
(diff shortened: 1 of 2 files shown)


test/lolex-test.js
```diff
@@ -637,6 +637,69 @@ describe("lolex", function () {
...
+            clock.tick(1990);
+            assert.equals(stub.callCount, 0);
+            clock.tick(20);
+            assert.equals(stub.callCount, 1);
+        });
+
+        it("is not influenced by forward system clock changes", function () {
+            var clock = this.clock;
+            var callback = function () {
+                clock.setSystemTime((new clock.Date()).getTime() - 1000);
```
fatso83: Any reason to use `(new clock.Date()).getTime()` instead of `clock.Date.now()`? It's not a blocking thing, so I'll merge this none the less, but still wondering.

kylefleming: I copied that line from the other tests that used setSystemTime, assuming there was a reason whoever wrote those tests wrote it that way.

PR #115, opened 2017-07-14
Add myself to authors list

benjamingr:
I just found out there is this file here - I assume it is used so people can contact authors of commits in case there is a problem - so adding my name and work address at Peer5 here. I also intend to actively help maintain lolex (as needed) as part of my job @ Peer5 :)


fatso83:
Thank you, but we update that file automatically as part of the [release process](https://github.com/sinonjs/lolex/blob/master/RELEASE.md). An active contributor is always welcome!

PR #117, opened 2017-07-16
Remove try/catch that captures exception

benjamingr:
I might not understand what that code does - but it looks like leftoff from previous versions of the code. 

I see no case in which there is an exception and it's not thrown - and the `if` seems redundant at which point we might as well avoid the try/catch.

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -286,26 +286,11 @@ function callTimer(clock, timer) {
         delete clock.timers[timer.id];
     }
 
-    try {
-        if (typeof timer.func === "function") {
-            timer.func.apply(null, timer.args);
-        } else {
-            /* eslint no-eval: "off" */
-            eval(timer.func);
-        }
-    } catch (e) {
-        exception = e;
-    }
-
-    if (!clock.timers[timer.id]) {
-        if (exception) {
-            throw exception;
-        }
-        return;
-    }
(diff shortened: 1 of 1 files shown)


fatso83:
Have a proper look into the history using `git blame` before doing that. Maybe we are missing a test to show the intent.

benjamingr:
@fatso83 the blame is 3 years ago for "working extraction from sinonjs" -  https://github.com/sinonjs/lolex/commit/dbd7d17029f5b258aa14422b74798c844de8bfab . 

If there is an exception - it is thrown in any case - and if there is no exception - it is not thrown anyway.

I think there _might_ have been some cleanup at some point in the past there - but it's not there anymore and that code is dead code now.

fatso83:
I went all the way back to the start of [that code in Sinon in 2010](https://github.com/sinonjs/sinon/blame/cc4940f35d2afca42190c690c5e379e464b3d1ad/src/sinon/util/fake_timers.js) (just click two times to be back at the first commit). Basically, back then, any exception was just swallowed. There really wasn't much to grasp from the evolution of that code.

And now, looking at the before/after of the code you are totally right - it had no function. No matter the `if` - any exception would be thrown either before or after the if, and there was no other code to be executed in between. 

Thanks for the cleanup :+1:

PR #119, opened 2017-07-16
Use added `keys`

benjamingr:
No point in polyfilling keys in one place and not using the polyfill later - either we remove the polyfill or we call it :)


fatso83:
Thanks!

Issue #116, opened 2017-07-16
`nextTick`


benjamingr:
I believe that an installation should also run `process.nextTick` calls if possible to simulate correct timer flow. 

Currently, lolex ignores `process.nextTick` calls which go out of order when testing Node calls :)

Fixing would look like:

 - detect if nextTick is present.
  - If it is, override it like `setTimeout` 
 - Whenever timers are run - always before the next timer - call the nextTick callbacks in order registered.

benjamingr:
Submitting a PR for this

fatso83:
Closed by #120. 

P.S. You might want to [check out this](https://help.github.com/articles/closing-issues-using-keywords/)

PR #120, opened 2017-07-16
Add support for process.nextTick

benjamingr:
Add support for nextTick with microtask semantics.

Microtask semantics follow Node behavior and ECMAScript's EnqueueJob
description. Microtasks are run between every "I/O" operation (timers
in this case).

This also paves the path for mocking promises. Since after this doing
and overriding Promise.prototype.then both promises supplied by the
platform and promises supplied by user libraries can be hooked on.


benjamingr:
Fixed by https://github.com/sinonjs/lolex/issues/116

Issue #122, opened 2017-07-18
Unable to use shouldAdvanceTime and advanceTimeDelta options.


royalpinto:
Env Details:
`lolex: 2.1.0`
`node: v6.9.2`

Following script throws an error:
```js
var lolex = require('lolex');
var clock = lolex.install({shouldAdvanceTime: true, advanceTimeDelta: 40});
setTimeout(() => console.log(1), 4000);
```

Error:
```bash
1
_stream_writable.js:396
  if (state.length === 0 && state.needDrain) {
           ^

TypeError: Cannot read property 'length' of undefined
    at onwriteDrain (_stream_writable.js:396:12)
    at afterWrite (_stream_writable.js:386:5)
    at runJobs (/Users/lohith.pinto/Development/stark/node_modules/lolex/src/lolex-src.js:183:18)
    at Object.tick (/Users/lohith.pinto/Development/stark/node_modules/lolex/src/lolex-src.js:550:9)
    at doIntervalTick (/Users/lohith.pinto/Development/stark/node_modules/lolex/src/lolex-src.js:411:11)
    at ontimeout (timers.js:365:14)
    at tryOnTimeout (timers.js:237:5)
    at Timer.listOnTimeout (timers.js:207:5)
```

What could be the issue ?

fatso83:
Thanks for your bug report. This was quite unfortunate.

@elad-nach, any ideas?

elad-nach:
@fatso83 I'm just looking into this as I'm trying to integrate 2.1.0 into sinon.

I'm also getting sporadic errors there - regardless to whether lolex is installed with `shouldAdvanceTime = true`. Errors are shown on mocha in the XHR test section on sinon.
I'm investigating in any case. will update once I nail it.

elad-nach:
@fatso83 @benjamingr 
could you please have a second look at the `nextTick()` feature that was merged? The `runJobs` was added on that PR.
see [here](https://github.com/angular/zone.js/issues/760) and [here](https://github.com/angular/zone.js/pull/762/files).

benjamingr:
The error makes sense - the user overrode `process.nextTick` and did not uninstall the timer - Node was trying to flush the ticks and lolex was there.

I'll see if I can hack around it (by mimicing nextTick better), the obvious fix is to advance the clock synchronously and then uninstall. 

A general way to reproduce would be:

```js
lolex.install({}); // 
startSomeAsyncNodeAction((err, data) => {
[...]

PR #123, opened 2017-07-18
Next tick parameters

benjamingr:
Lesson learned, test added, code fixed. 

Fixes https://github.com/sinonjs/lolex/issues/122


test/lolex-test.js
```diff
@@ -2053,6 +2053,17 @@ describe("lolex", function () {
                 clock.runAll();
                 assert(!called);
             });
+
+            it("passes arguments when installed - GitHub#122", function () {
+                var clock = this.clock = lolex.install();
```
fatso83: why assign to `this.clock`? It is never used in the code. It seems quite sufficient to only set the local variable.

benjamingr: Oh, I meant to do that and use it to uninstall - and then uninstalled myself. Fixing.

benjamingr: Fixed (removed uninstall call)

elad-nach: i think he meant on line 2058
`var clock = this.clock = lolex.install();`
could be changed to:
`var clock = lolex.install();`

benjamingr: @elad-nach - the problem was it also `clock.uninstall()`d, `var clock` is so I don't have to type `this` before it and there is a `afterEach` hook that uninstalls already. The tests above this one show the pattern.

elad-nach:
this is somehow still broken.

I'm trying to integrate 2.1.1 to sinon and the nextTick functionality somehow is still breaking something else on the `test-headless`.

I'm trying to find out what exactly. 

Commenting it out the relevant nextTick lines resolves the problem.

fatso83:
@elad-nach Using Sinon as an integration test is a pretty good idea in any case, which we perhaps might consider in any case. Thanks for doing it and uncovering stuff like this in any case. I'll try to do it myself the next time a patch comes in.

benjamingr:
> I'm trying to integrate 2.1.1 to sinon and the nextTick functionality somehow is still breaking something else on the test-headless.

We can detect if the nextTick was run from Node code and not our own and run node's own microticks - can you try adding a line to run the original nextTick if the call originated in Node and see if that helps (by checking the current stack trace).

elad-nach:
i fail to understand - what's the difference? nextTick is a node functionality.
and a couple of questions at that:
1. how was process.nextTick() handled before this change?
2. looking at the implementation - runJobs is equivalent to nodeJS' nextTick()? nextTick means that the callbacks should be executed after the user-code has finished executing - how do you defer the execution to that stage?

benjamingr:
> how was process.nextTick() handled before this change? would it just hang?

Lolex would not handle `nextTick` at all - it would be like an "uncovered timer"

> looking at the implementation - runJobs is equivalent to nodeJS' nextTick()? 

No, `nextTick` is, `runJobs` is equivalent to "RunJobs" from here: http://ecma-international.org/ecma-262/8.0/#sec-runjobs

Node does this here:

 - First, a NextTickQueue is defined (I used an array for this here, since it's a testing library and a fast queue dependency isn't worth it). https://github.com/nodejs/node/blob/master/lib/internal/process/next_tick.js#L13
 - Doing a `process.nextTick` does https://github.com/nodejs/node/blob/master/lib/internal/process/next_tick.js#L107-L117 (pushes into the queue, albeit cleverly).
 - Which in turn has a hook in node.cc which runs all the microtasks after a tick. Here: https://github.com/nodejs/node/blob/master/src/node.cc#L1230-L1253 - as `runMicrotasks`. 
 - This hooks `RunMicrotasks` which is a V8 feature: https://github.com/nodejs/node/blob/master/src/node.cc#L1358 

> nextTick means that the callbacks should be executed after the user-code has finished executing - how do you defer the execution to that stage?

Same way as timers - I don't run them at all - but any timer ticking does. The only "special" thing is that I have to runJobs between any two opportunity for timers - since microticks `nextTick`d during a timer have to happen before the following timer (except in a specific edge case I'd rather not get into). 

```js
var clock = lolex.install();
process.nextTick(() => console.log("A"));
clock.runAll(); // this runs the nextTick, like a timer. 
```

Basically, the important part is that like other timers in lolex - we are able to tell it "progress asynchronous tasks".

elad-nach:
as far as I understand - nextTick has nothing to do with time ticking; why invoke it from within the timer context when it has to satisfy other event loop state constraints.

Before the nextTick implementation on lolex - process.nextTick() was handled by node correctly:
```
describe('a test', function () {
    'use strict';
    const llx = require('lolex');
    it('should do the next tick', (done) => {
        llx.install();

        process.nextTick(() => {
            console.log('1');
            done();
        });
    })
})
```
this gets executed with lolex 1.6.0 and node 6.9.1 so I guess I'm missing something?

benjamingr:
@elad-nach the whole point of lolex is to convert test code that is supposed to be asynchronous into more controllable synchronous code - however: it failed with code using nextTick - not only can the above code sample be written completely synchronously in lolex 2 - but requires an asynchronous test in lolex 1. 

Not to mention things that never worked (like timers inside `nextTick`). 

`nextTick` isn't about "time ticking" - but it has to happen in order to scheduler code relying on timers correctly. Once I get the one other edge case (promises) - we can write tests for complex asynchronous code synchronously with lolex - which would be great.

(setImmediate doesn't have anything to do with time ticking either for what it's worth)

elad-nach:
timers inside `nextTick` wouldn't execute because timers have to be triggered as an interrupt, which was not possible when using lolex without utilising the original system timers for interrupts (that's why I implemented `shouldAdvanceTime` feature that piggybacks over the system timers' advancements), not because of `nextTick`'s behaviour.

Basically without explicitly calling `clock.runAll()` or `shouldAdvanceTime=true` (default is `false`) or calling `tick()` this implementation will never get called.

Anyone that will want to use a fake timer in an environment where `process.nextTick()` is called, will now have to manually `tick()` it, call the `runAll` or install with `shouldAdvanceTime=true`.

benjamingr:
> timers inside nextTick wouldn't execute because timers have to be triggered as an interrupt, which was not possible when using lolex without utilising the original system timers for interrupts 

Timers inside `nextTick` wouldn't execute because `nextTick` wouldn't happen synchronously when you choose - but it was left to the platform (unlike things like setImmediate). 

> Basically without explicitly calling clock.runAll() or shouldAdvanceTime=true (default is false) or calling tick() this implementation will never get called.

Exactly, that's the point, you get to synchronize your flow.

>  Anyone that will want to use a fake timer in an environment where process.nextTick() is called, will now have to manually tick() it, call the runAll or install with shouldAdvanceTime=true.

Anyone who wants to use a fake timer which defaults to not progressing time - would not expect time to progress. Calling `nextTick` is _exactly_ like calling `setImmediate` (sans i/o order) and it made no sense to support the one and not another.

The point of lolex is to fake timers and to be able to break application tests into discrete synchronization points. Without nextTick - that never really worked for node (but worked fine for browsers).

What we _can_ do however is still run `nextTick` calls that are not a part of user code (like asynchronous code).

elad-nach:
I agree with you on some of the points you've made but under real life conditions unless you execute every `nextTick` call synchronously (which might break other things for some users due to execution context) you won't be able to know when to `runAll` from within the test context because you don't know when `nextTick` is called.

Now what happens to sinon users that used the fake timers? will they all have to revise their tests as of such?

benjamingr:
> you don't know when nextTick is called.

I think you can - it's called when you run your clock or tick it (we override nextTick on `.install`) - that was the whole point of the PR :)

Issue #67, opened 2016-06-09
Fake setTimeout should "fail" on too large delays


fiddur:
Both browsers and Node.js handles timers with int32 and overflows on values above 2147483647.  It would be good if lolex in some way fails in that case too.

Background:

I [read about](https://nodejs.org/api/timers.html#timers_settimeout_callback_delay_arg) the maximum delay on `setTimeout`, and prepared a `farTimeout` to just recursively wrap the  `setTimeout` with max timeout multiple times.  But, using test first approach, I noticed that the `setTimeout` handled 90 days without problem, even though the docs said max ~25 days.  Digging deeper, I realized that the sinon.useFakeTimers gave me a `setTimeout` that handled larger numbers than Node.js `setTimeout`.

Suggested solution:

I think something like this in `addTimer` function would make it behave more like the actual `setTimeout`:

``` javascript
const maxTimeout = 2147483647
if (timer.delay && timer.delay > maxTimeout) {
    timer.delay = (timer.delay + maxTimeout) % (maxTimeout * 2) - maxTimeout
}
```

…not making a pull request right now though since this was quite quickly thrown together; could make one if you think this is a good solution.

fatso83:
Seems reasonable. But the if is unneeded, so the math could be simplified to

```javascript
// overflow if timer > maxTimeout
timer.delay = timer.delay % maxTimeout;
}
```

fiddur:
I placed it in `addTimer` rather than `setTimeout` since there are similar limitations to other timers, but in `addTimer` there isn't always a `delay`?

The `+ max … - max` is to mimic the int32 overflow, that will sometimes render a negative value (making the timer trigger immediately).

fatso83:
There are "timers" like `setImmediate` and `nextTick` that does not take arguments, so that might be why.

fatso83:
Was there a PR coming?

fiddur:
I'm sorry, my vacation got in the way!  I'll get to it this week I hope!

fatso83:
New summer, new hopes :-)

benjamingr:
There are a few edge cases: 

Node does:
> When delay is larger than 2147483647 or less than 1, the delay will be set to 1.

Timers API say:
> If timeout is less than 0, then set timeout to 0.

We don't handle this case at the moment - can PR to handle. We do handle it not being passed.

The spec itself does _not_ require that the timeout is under 2**32 nor does it require that the timers are overflown in a particular way.

MDN claims:
[...]

PR #124, opened 2017-07-19
issue #67 - overflowing the timer correctly according to int32_t

elad-nach:
i did not change the logic for handling `setTimeout` with a value which is smaller than 1. I think this is redundant since some users might be using `setTimeout` with a value of 0, then `tick()`ing it might result in an unexpected sequence of timers being executed. There's one test that demonstrates this (`"calls the given callback before setTimeout"`).
not sure if we need to change this edge case.

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -2,6 +2,7 @@
 
 var userAgent = global.navigator && global.navigator.userAgent;
 var isRunningInIE = userAgent && userAgent.indexOf("MSIE ") > -1;
+var maxTimeout = Math.pow(2, 31) - 1; //see https://heycam.github.io/webidl/#abstract-opdef-converttoint
 
 // Make properties writable in IE, as per
 // http://www.adequatelygood.com/Replacing-setTimeout-Globally.html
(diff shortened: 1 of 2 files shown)


benjamingr:
Mirroring my comment from before - I think we need to align with https://heycam.github.io/webidl/#abstract-opdef-converttoint for the timer.

src/lolex-src.js
```diff
@@ -483,6 +484,7 @@ function createClock(now, loopLimit) {
         });
     };
     clock.setInterval = function setInterval(func, timeout) {
+        timeout = timeout > maxTimeout ? 1 : timeout;
```
benjamingr: I'd rather have all the bounds checking logic once in `addTimer`

test/lolex-test.js
```diff
@@ -53,6 +53,28 @@ describe("issue #73", function () {
     });
 });
 
+describe("issue #67", function () {
+    // see https://nodejs.org/api/timers.html
+    it("should overflow to 1 on very big timeouts", function () {
```
benjamingr: This is good, I would prefer more tests :)

elad-nach:
as I mentioned in the issue comment - I understand the IDL reference as a reference but I'm not sure about your request - are you asking to write the integer conversion as specified in the IDL?

benjamingr:
@elad-nach I'm wondering if the correct thing would be to implement the IDL (which is what's discussed in the original issue) or not. I can't get Chrome to act according to the spec exactly - so I'm a little torn on what behavior we want.

benjamingr:
In any way, I think we should land this - as it's better to specify this behavior in a useful way and to debate correctness after :)

PR #127, opened 2017-07-25
issue #126 - do not attach `nextTick` by default

### LINKED ISSUE #126: Faked `nextTick` causes Node.js 8.2.0 to hang
https://github.com/novemberborn/ksuid/commit/b4564c5d52835e1ce479dafe9cea938b1a738dfd hangs when tests are run in Node.js 8.2.0, but is fine in 8.1.4.

The fix is to not fake `nextTick`: https://github.com/novemberborn/ksuid/commit/7239faa5d166b7a241eb78c688d0c0191f72ed06

Tests are run using AVA, which does not directly uses `process.nextTick()`. I'm not sure why it the hangs occur with Node.js 8.2.0 either. It does seem that faking `nextTick` by default may result in unexpected behavior.

elad-nach:
resolves #126 
- does not fake process.nextTick by default
- added .idea folder to .gitignore
- added missing clock teardowns after installs

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -142,7 +142,7 @@ Parameter | Type | Default | Description
 --------- | ---- | ------- | ------------
 `config.target`| Object | global | installs lolex onto the specified target context 
 `config.now` | Number/Date | 0 | installs lolex with the specified unix epoch 
-`config.toFake` | String[] | [] | an array with explicit function names to hijack. You can pick from `setTimeout`, `clearTimeout`, `setImmediate`, `clearImmediate`,`setInterval`, `clearInterval`, and `Date`. E.g. `lolex.install({ toFake: ["setTimeout","clearTimeout"]})`
+`config.toFake` | String[] | ["setTimeout", "clearTimeout", "setImmediate", "clearImmediate","setInterval", "clearInterval", "Date"] | an array with explicit function names to hijack. *When not set, lolex will automatically fake all methods **except** `nextTick`* e.g., `lolex.install({ toFake: ["setTimeout","nextTick"]})` will fake only `setTimeout` and `nextTick`
 `config.loopLimit` | Number | 1000 | the maximum number of timers that will be run when calling runAll()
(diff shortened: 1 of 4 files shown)


src/lolex-src.js
```diff
@@ -689,7 +689,8 @@ exports.install = function install(config) {
     clock.methods = config.toFake || [];
 
     if (clock.methods.length === 0) {
-        clock.methods = keys(timers);
+        // do not fake nextTick by default - GitHub#126
+        clock.methods = keys(timers).filter(function (key) {return key !== "nextTick";});
```
fatso83: This means we stub all the other timers by default, but this is not reflected in the docs, which erronously says the default is an empty array. Care to fix the docs? It's the line on `config.toFake` in the README.

elad-nach: done

benjamingr:
Thanks

fatso83:
Released as 2.1.2 on NPM.

Readme.md
```diff
@@ -142,7 +142,7 @@ Parameter | Type | Default | Description
 --------- | ---- | ------- | ------------
 `config.target`| Object | global | installs lolex onto the specified target context 
 `config.now` | Number/Date | 0 | installs lolex with the specified unix epoch 
-`config.toFake` | String[] | [] | an array with explicit function names to hijack. You can pick from `setTimeout`, `clearTimeout`, `setImmediate`, `clearImmediate`,`setInterval`, `clearInterval`, and `Date`. E.g. `lolex.install({ toFake: ["setTimeout","clearTimeout"]})`
+`config.toFake` | String[] | ["setTimeout", "clearTimeout", "setImmediate", "clearImmediate","setInterval", "clearInterval", "Date"] | an array with explicit function names to hijack. *When not set, lolex will automatically fake all methods **except** `nextTick`* e.g., `lolex.install({ toFake: ["setTimeout","nextTick"]})` will fake only `setTimeout` and `nextTick`
```
SimenB: is this missing `hrtime`?

benjamingr: @SimenB hrtime and performance.now are always mocked (mocked time) IIRC

PR #137, opened 2017-11-02
Add support for requestAnimationFrame

mAAdhaTTah:
I could really use a mocked `requestAnimationFrame` for some testing I'm doing, so I thought I'd take a crack at adding it here. I opted to just tick a rAF every 16ms, so the callbacks can just be run with `clock.tick(16)`. I think this mostly works / does what's needed, but I'm not sure of a few things, like bootstrapping this only in a browser or passing in the correct value for the callback arg. Let me know if I'm on the right track.


mAAdhaTTah:
Added some tests to ensure callbacks are scheduled together in the next frame. Also added a `runToFrame` method to the clock to run the timer through the next frame. All of these have tests. If this looks like a good approach overall, I can add some docs.

Let me know what you think!

mAAdhaTTah:
Updated from master after merge of #139.

fatso83:
LGTM! Mind rebasing it instead of the merge :grimacing: ?

mAAdhaTTah:
Sure! I'll add some docs too.

mAAdhaTTah:
Alright, rebased, squashed, and updated the docs.

fatso83:
Published as 2.2.0!

PR #140, opened 2017-11-07
Update Mocha to 4.0.1

fatso83:
Avoid breaking bitkeeper test


fatso83:
Argh, failed. Needs mantoni/mochify.js#160 to be merged, as well as updating `mochify` after that.

PR #145, opened 2017-11-22
bugfix for a setTimeout() or setSystemTime() within a nextTick() call.

rogierschouten:
```javascript
var lolex = require('lolex');

var clock = lolex.install({ 
	now: 0, 
	toFake: [
		"setTimeout", "clearTimeout", "setImmediate", "clearImmediate", "setInterval", "clearInterval", "Date", "nextTick"
	]
});

var callbackCalled = false;
process.nextTick(function() {
	setTimeout(function() {
			callbackCalled = true;
	}, 50);
});

// BUG: infinite loop!
while (!callbackCalled) {
	clock.tick(60); // note more than the delay of the setTimeout() call
}
```


benjamingr:
Why did you delete the `package.lock` @rogierschouten?

rogierschouten:
@benjamingr  Oh wait I thought I introduced the package-lock.json. I'll reintroduce it. 

But I just also found a second bug relating to this, and I'm losing track of how to fix this.

The clock.tick() function contains this code:
```
        runJobs(clock);
        clock.duringTick = false;
        updateHrTime(tickTo);
        clock.now = tickTo;
```

Notice how `clock.now` is only updated after the `runJobs()`.     I just created another setTimeout() within a nextTick() and it doesn't go off - because the statement `clock.now = tickTo` jumps right over the newly added timer time.

Let me chew for a moment on this

rogierschouten:
@benjamingr  I think I have all the nextTick() related bugs now, I also added some tests that fail before my change and that are ok after my change. Hope this helps.

benjamingr:
Basically, `runJobs` needs to happen at every two possible cases where timers get to run.

rogierschouten:
Yes, and also, the local variables in `clock.tick()` need to remain compensated for any `setSystemTime()` calls that have come while running timers or jobs.

rogierschouten:
Is this ok to get merged or do you need more from me?

fatso83:
Squashing the changes cleaned up the history, so all good.

fatso83:
2.3.1 released. :smile:

PR #150, opened 2017-12-24
Add missing functions in default `toFake`

SimenB:
I _think_ this is correct.

It would be great if I was able to pass `true` or `'all'` or something, so that I can get all future added timers for free even though I opt in to fake `nextTick`.

It seems like methods are only added if  they exist in the original global, even though they are specified in `toFake`. Is this understanding correct? Should it be clarified in the docs either way?

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -139,7 +139,7 @@ Parameter | Type | Default | Description
 --------- | ---- | ------- | ------------
 `config.target`| Object | global | installs lolex onto the specified target context
 `config.now` | Number/Date | 0 | installs lolex with the specified unix epoch
-`config.toFake` | String[] | ["setTimeout", "clearTimeout", "setImmediate", "clearImmediate","setInterval", "clearInterval", "Date"] | an array with explicit function names to hijack. *When not set, lolex will automatically fake all methods **except** `nextTick`* e.g., `lolex.install({ toFake: ["setTimeout","nextTick"]})` will fake only `setTimeout` and `nextTick`
+`config.toFake` | String[] | ["setTimeout", "clearTimeout", "setImmediate", "clearImmediate","setInterval", "clearInterval", "Date", "requestAnimationFrame", "cancelAnimationFrame", "hrtime"] | an array with explicit function names to hijack. *When not set, lolex will automatically fake all methods **except** `nextTick`* e.g., `lolex.install({ toFake: ["setTimeout","nextTick"]})` will fake only `setTimeout` and `nextTick`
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -139,7 +139,7 @@ Parameter | Type | Default | Description
 --------- | ---- | ------- | ------------
 `config.target`| Object | global | installs lolex onto the specified target context
 `config.now` | Number/Date | 0 | installs lolex with the specified unix epoch
-`config.toFake` | String[] | ["setTimeout", "clearTimeout", "setImmediate", "clearImmediate","setInterval", "clearInterval", "Date"] | an array with explicit function names to hijack. *When not set, lolex will automatically fake all methods **except** `nextTick`* e.g., `lolex.install({ toFake: ["setTimeout","nextTick"]})` will fake only `setTimeout` and `nextTick`
+`config.toFake` | String[] | ["setTimeout", "clearTimeout", "setImmediate", "clearImmediate","setInterval", "clearInterval", "Date", "requestAnimationFrame", "cancelAnimationFrame", "hrtime"] | an array with explicit function names to hijack. *When not set, lolex will automatically fake all methods **except** `nextTick`* e.g., `lolex.install({ toFake: ["setTimeout","nextTick"]})` will fake only `setTimeout` and `nextTick`
```
SimenB: is this missing `performance.now`? How is it specified?

fatso83: I don't have opportunity to delve into the code right now (vacation in remote location, briefly smartphone, maybe Ben has a chance?), but I can look into it next week.

benjamingr: What do you mean how is it specified?

SimenB: You say "hrtime" and "nextTick", not "process.hrtime" or "process.nextTick". Do you just say "now" for `performance.now()`? It would be consistent, but look weird.

benjamingr: `performance.now` should be used, mostly because only `process` has `.hrtime` and `nextTick` but `now` can be `Date.now()` or `performance.now` - although we should probably change it to `process.hrtime` `process.nextTick` and `performance.now` to not be confusing.

benjamingr:
@fatso83

Readme.md
fatso83: I agree that we should change it to `performance.now` to be precise, but that would require a code change and a major bump, since it is a api change. If you are not too keen on looking into that in this commit due to scope creep, Simen, I totally understand, and we could do that in a later PR. So basically, this is good to merge, but I'll just wait for the reply.

SimenB: Merge away 🙂

PR #153, opened 2018-01-17
Fix typo in README

jlfwong:


jlfwong:
I'm not sure what's up with CI, but it seems suspect that it's failing as the result of a 1 character change to the README :)

benjamingr:
@jlfwong the CI passed (bitHound is unrelated). Thank you for your help!

fatso83:
Thanks!

Issue #146, opened 2017-12-23
Create fake timers based on passed in object


SimenB:
Bad title, so I'll try to describe my use case.

If I create a JSDOM instance in node, it has `requestAnimationFrame`, but lolex is unable to discover this (unless I use some fancy `vm` code) as is inspects `global` at `require`-time. This means that I have to use `lolex` inside of the script I execute in JSDOM, instead of managing it from the outside by controlling the injected globals. I need this as I'm looking to use Lolex directly in Jest.

I'm thinking something like `const clockFactory = lolex.createClockFactoryFrom(global)` or something, which can be separately used later in for instance `const clock = clockFactory.install({ 
target: myTarget });`

jdalton:
For context, @SimenB is a Facebook dev for the Jest test framework. Jest is looking to adopt lolex for their mocks to simplify their support and defer to a specialized lib like sinonjs/lolex.

benjamingr:
Sure @jdalton @SimenB I think we can do that. 

The way I see it - we'd either pass the object to `.install` or expose a new method that does it to `global`.

Alternatively, we can expose a factory on `lolex`:

 - Wrap the entire code in `lolex-src.js` in a factory function
 - Invoke that when we export lolex with `global` from the outside (from browserify)
 - Expose the factory a second time as `lolex.initializeFromGlobal`.

That should let you pass `global` in. As an alternatively, jest could create a new clock and then override the globals itself - but that doesn't sound ergonomic and I think it's fine for us to expose it (especially since the JSDOM in Node use case is entirely legitimate and should be supported IMO)

@fatso83 if this sounds good to you - I can open a PR for this if you'd like - shouldn't be too much work to add.

SimenB:
>As an alternatively, jest could create a new clock and then override the globals itself

Yeah, that's the alternative, but I was hoping to offload all of that logic onto Lolex.

One issue with manually setting the globals is that `set{Timeout,Interval,Immediate}` returns `Timer` instances with an `unref` function in Node which we don't want with JSDOM. I assume from the docs that Lolex does the right thing:tm: based on globals, which breaks in the JSDOM case.

<sup>Just to clarify, I'm a collaborator on Jest, but I don't work for FB 🙂</sup>

benjamingr:
@SimenB 

> One issue with manually setting the globals is that set{Timeout,Interval,Immediate} returns Timer instances with an unref function in Node which we don't want with JSDOM

[...]

PR #157, opened 2018-02-15
Allow configuring if lolex should return DOM or node timers

benjamingr:
Fixes (possibly) https://github.com/sinonjs/lolex/issues/146

CC @SimenB - let me know if this addresses your environment configuration issue with `global` or if a more general workaround (passing the library) is still required - this just felt less flakey to me.

CC @fatso83

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -261,6 +261,10 @@ setSystemTime().
 Restores the original methods on the `target` that was passed to
 `lolex.install`, or the native timers if no `target` was given.
 
+### `lolex.setTimersReturnsObjects(boolean)`
+
+Changes whether or not timers should return objects (like Node native timers) or integer IDs (like DOM timers).
+
 ### `Date`
 
 Implements the `Date` object but using the clock to provide the correct time.
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -738,6 +738,12 @@ function createClock(start, loopLimit) {
 }
 exports.createClock = createClock;
 
+exports.setTimersReturnsObjects = function(value) {
+    var oldValue = addTimerReturnsObject;
+    addTimerReturnsObject = value;
+    return oldValue;
+};
+
 /**
  * @param config {Object} optional config
  * @param config.target {Object} the target to install timers in (default `window`)
--- a/test/lolex-test.js
+++ b/test/lolex-test.js
@@ -1726,6 +1726,22 @@ describe("lolex", function () {
             }
         });
 
+        it("Configuring the type of timers", function () {
+            this.clock = lolex.install();
+            var stub = sinon.stub();
+
+            var oldValue = lolex.setTimersReturnsObjects(true);
+            var to = setTimeout(stub, 1000);
+            assert.isNumber(to.id);
+            assert.isFunction(to.ref);
+            assert.isFunction(to.unref);
+            var returnValue = lolex.setTimersReturnsObjects(false);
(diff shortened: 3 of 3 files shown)


SimenB:
This won't be enough for Jest's use case as for instance `requestAnimationFrame` is not on `global`. It will be on `target`, though. It will also incorrectly detect `process.nextTick` even though it won't be available in the runtime of tests

benjamingr:
Interesting, I'm still leaning towards exposing the detection outside as configuration rather than accepting a global.

I figured jest wouldn't be interested in things like whether or not `requestAnimationFrame` is shimmed by default since it picks which timers to shim itself. Would exposing the rest of the flags help?

SimenB:
Jest cares that it is shimmed at all, which it won't be without having lolex do it's detections either on `target` passed to `lolex.install` or through some other way of inspecting an object I give it.

if `lolex.install({ target })` did all of its sniffing for existent timer functions on `target` it would work. Now `performancePresent ` will be `false` meaning `performance.now` will be gone, whilst `hrtimePresent` will be true, menaing it's exposed even if it shouldn't be

benjamingr:
@SimenB what about passing `Object.keys(window)` for the JSDOM window or `Object.keys(global)` in Node as `toFake` to `lolex.install`? 

If you want hrtime you can also pass `Object.keys(global).concat(Object.keys(process)`

SimenB:
That might work, yeah. Same with `performance` I guess?

benjamingr:
@SimenB right, but not with whether or not timers should or shouldn't return objects - which requires this PR. Would you be OK with that?

I'm still willing to allow passing target more "aggressively" but I'm not really sure that's particularly good API behavior and this lets us do things explicitly.

SimenB:
Doesn't work to pass any old string to `toFake` because of this line: https://github.com/sinonjs/lolex/blob/444c9384ace3853761d76aa31fa4ac3648247c9d/src/lolex-src.js#L409

```
    TypeError: Cannot set property 'hadOwnProperty' of undefined

      at hijackMethod (node_modules/lolex/src/lolex-src.js:409:34)
```

It's also impossible to fake `performance.now` as lolex thinks it doesn't exist. Unless I'm doing something weird, it seems impossible to make lolex detect the correct globals when it's self-detecting at `require`-time instead of inspecting some passed in object.

fatso83:
Any thoughts on how to proceed with the [issues @SimenB mentions](https://github.com/sinonjs/lolex/pull/157#issuecomment-370222256)? Is that a separate issue?

SimenB:
Any news here? We're gearing up for Jest 23, and switching will probably be pretty breaking, so I'd love to be able to include Lolex in that release 🙏 

I'm happy to put together a PR if we can figure out the correct API. The issue boils down to the fact that `lolex` inspects globals when it's `require`d, but Jest has a separate global object which is used in tests, and that's the thing that needs to be inspected.

benjamingr:
> I'm happy to put together a PR if we can figure out the correct API. The issue boils down to the fact that lolex inspects globals when it's required, but Jest has a separate global object which is used in tests, and that's the thing that needs to be inspected.

To be honest I'm not really 100% sure what a solution would be - so a PR would be great :)

SimenB:
@benjamingr cool, opened up #164

benjamingr:
Thanks, let's continue the discussion there, I'm +1 on landing the changes in time for the next jest major

Issue #136, opened 2017-10-28
window.performance.mark becomes undefined after timer install


oliviertassinari:
I have noticed the following behavior in the browser:
```js
import { useFakeTimers } from 'sinon';

console.log('1', window.performance.mark) // '1', function mark() { ... }
clock = useFakeTimers();
console.log('2', window.performance.mark) // '2', undefined
```

`window.performance` became a function after calling `useFakeTimers`. I believe that the issue is coming from this line:
https://github.com/sinonjs/lolex/blob/c08ff362a47e9786bc62da6c37175838bb7bd21e/src/lolex-src.js#L406-L408

One simple workaround is the following:
```diff
-clock = useFakeTimers();
+clock = useFakeTimers({
+  toFake: ["setTimeout", "clearTimeout", "setImmediate", "clearImmediate","setInterval", + "clearInterval", "Date"]
+});
```

oliviertassinari:
I had to do some digging, the error on my side was coming from this line: https://github.com/facebook/react/blob/v16.0.0/src/renderers/shared/fiber/ReactDebugFiberPerf.js#L81.

fatso83:
Verified. Here is the test with a browserified transpiled output that can be copy-pasted directly into Chrome for verification. Tested with `master` (6476935).

[test.js.txt](https://github.com/sinonjs/lolex/files/1426811/test.js.txt)
[transpiled.js.txt](https://github.com/sinonjs/lolex/files/1426812/transpiled.js.txt)

benjamingr:
@oliviertassinari wanna take a shot at a PR at this? Otherwise I can probably take a stab

oliviertassinari:
@benjamingr I'm already pretty busy with Material-UI. Sorry, it's not a priority for me.

benjamingr:
@oliviertassinari no need to justify or apologize - it's entirely fine if it's not a priority for you - I was merely offering :)

benjamingr:
@fatso83 I still can't assign myself to issues by the way :D I'll take a stab next week hopefully

medialwerk:
Is there anyone working at this issue at the moment? :-)

benjamingr:
sorry, completely forgot about it - can take a look early next week

medialwerk:
Thank you very much!

risinglf:
Hi guys, news about this issue? thanks

fatso83:
[...]

PR #160, opened 2018-04-05
Fix that performance.mark is undefined after timer install

CoderK:
## Related issue
https://github.com/sinonjs/lolex/issues/136

## Why

performance.mark method is not an own property.

```javascript
console.log(Object.getOwnPropertyNames(performance)) // [], it's empty.
```

It belongs to Performance.prototype.
Because all properties in performance._\_proto__ are not clock.performance's own properties, 

```javascript
clock.performance = Object.create(global.performance);
```
they couldn't pass the following conditional statements.

```javascript
target[method] = function () {
    return clock[method].apply(clock, arguments);
};

for (prop in clock[method]) {
    if (clock[method].hasOwnProperty(prop)) {
        target[method][prop] = clock[method][prop];
    }
}
```

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -710,6 +710,15 @@ function createClock(start, loopLimit) {
 
     if (performancePresent) {
         clock.performance = Object.create(global.performance);
+
+        var proto = Performance.prototype;
+
+        Object
+            .getOwnPropertyNames(global.Performance.prototype)
+            .forEach(function (name) {
+                clock.performance[name] = proto[name];
+            });
+
         clock.performance.now = function lolexNow() {
             return clock.hrNow;
         };
--- a/test/lolex-test.js
+++ b/test/lolex-test.js
@@ -1833,6 +1833,26 @@ describe("lolex", function () {
                 this.clock.uninstall();
                 assert.same(performance.now, oldNow);
             });
+
+            it("keeps original performance's all prototype methods after being installed", function () {
+                // In Phantom.js environment, Performance.prototype has only "now" method.
+                // For testing, some stub functions need to be assigned.
+                Performance.prototype.someFunc1 = function () {};
+                Performance.prototype.someFunc2 = function () {};
+                Performance.prototype.someFunc3 = function () {};
+
+                this.clock = lolex.install();
+
+                assert.isFunction(performance.someFunc1);
+                assert.isFunction(performance.someFunc2);
+                assert.isFunction(performance.someFunc3);
+
+                this.clock.uninstall();
+
+                delete Performance.prototype.someFunc1;
+                delete Performance.prototype.someFunc2;
+                delete Performance.prototype.someFunc3;
+            });
         }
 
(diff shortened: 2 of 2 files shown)


src/lolex-src.js
```diff
@@ -710,6 +710,15 @@ function createClock(start, loopLimit) {
 
     if (performancePresent) {
         clock.performance = Object.create(global.performance);
```
CoderK: I originally wanna write code like this,
```javascript
clock.performance = Object.create(global.Performance.prototype);
```

In Phantom.js environment, prototype's properties that be assigned by developer didn't be copied for unknown reason.

test/lolex-test.js
```diff
@@ -1833,6 +1833,26 @@ describe("lolex", function () {
                 this.clock.uninstall();
                 assert.same(performance.now, oldNow);
             });
+
+            it("keeps original performance's all prototype methods after being installed", function () {
```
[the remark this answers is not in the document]
CoderK: I think you want me to write test like this, right?

‘’’javascript
It keeps performance.mark after installed.
‘’’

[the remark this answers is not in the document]
fatso83: Yes (although a typo: 'should keep'), and possibly change the other test name to 
`"should not alter the global performance properties and methods"`

fatso83:
Handling this now.

fatso83:
Merged as e745324. Thank you!

CoderK:
I have been busy these days and have not been able to modify the code. Thank you!

fatso83:
@CoderK After merging this [the builds started failing](https://travis-ci.org/sinonjs/lolex/jobs/367234969#L2533h). Could you look into fixing it? I think it just requires some defensive checks.

CoderK:
Oops!! Ok, I see. #169 is already doing about this, but I am gonna review it too.

CoderK:
I'm sorry. It's my mistake. I couldn't think that some Performance's properties are read-only in some browsers. I didn't know there was an integration test. ㅜ-ㅜ. The issue is likely to be resolved at #169

src/lolex-src.js
```diff
@@ -710,6 +710,15 @@ function createClock(start, loopLimit) {
...
+
+        var proto = Performance.prototype;
+
+        Object
+            .getOwnPropertyNames(global.Performance.prototype)
+            .forEach(function (name) {
+                clock.performance[name] = proto[name];
```
fatso83: @CoderK I just changed this line to fix #200 in my PR #203. Do you see any reason not to replace the methods with stubs? Calling performance.mark() and such started failing after this (PR 160).

CoderK: @fatso83 
I had no the special reason not to replace them with stubs. I just thought that if I have no a specific reason, I should preserve original function as much as possible when mock object is created.

fatso83: Yeah, it seemed like calling those methods failed when not in the original object. At least in some browsers.

Issue #158, opened 2018-02-15
Make performance work with Node


benjamingr:
See https://nodejs.org/api/perf_hooks.html

I can take a look or mentor someone who wants to look into this

PR #162, opened 2018-04-21
Upgrade Mochify to v5.6

mantoni:
It turned out that some properties on the `Performance` property are not writable in Chrome. I'm a little surprised this wasn't reported by users yet.

Even though the Travis build fails due to puppeteer installation issues, you can verify locally that this change makes upgrading to Mochify v5 possible.

Maybe related: #158

Update: Running `npm run test-cloud` also works, verifying that tests pass in IE 9, 10, 11 and Firefox.

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -738,7 +738,9 @@ function withGlobal(_global) {
             Object
                 .getOwnPropertyNames(_global.Performance.prototype)
                 .forEach(function (name) {
-                    clock.performance[name] = proto[name];
+                    if (Object.getOwnPropertyDescriptor(proto, name).writable) {
+                        clock.performance[name] = proto[name];
+                    }
                 });
 
             clock.performance.now = function lolexNow() {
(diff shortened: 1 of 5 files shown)


fatso83:
It's great that this enables us to run Mochify, but if it will consistently break our tests I don't see why we should merge it atm? I am assuming Mochify wraps Puppeteer, and currently, [Travis requires Puppeteer to be called in a special way](https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-on-travis-ci) to disable the sandbox. I guess we need to wait for either Travis to update its kernel versions, or we need Mochify to support passing in a flag that can be passed to puppeteer (which we would put in our Travis config) before we can merge this.

mantoni:
Ok, rebased against master and changed the travis config to match what we did in the Sinon project. This means we don't need the `script/ci-test.sh` script anymore.

If you click through the builds you will see that Node 4 & Node 6 only run the node tests while Node 8 runs the linter and Mochify with headless Chrome.

From my side this is good to merge.

mantoni:
This fixes #167 and #170.

mantoni:
@fatso83 Thanks! Can I ask you to cut a release for this? I don't have the required tools installed.

fatso83:
@mantoni It was out as 2.4.2 a few minutes later. Just took a while, as I was working from home, and my Cygwin install hadn't been working with Lolex for a couple of years, so took some time to get everything running :-)

mantoni:
@fatso83 Awsome! Thanks a lot. Tested it in a project that broke and it's now all working great again 😃

src/lolex-src.js
```diff
@@ -738,7 +738,9 @@ function withGlobal(_global) {
             Object
                 .getOwnPropertyNames(_global.Performance.prototype)
                 .forEach(function (name) {
-                    clock.performance[name] = proto[name];
+                    if (Object.getOwnPropertyDescriptor(proto, name).writable) {
+                        clock.performance[name] = proto[name];
```
fatso83: Changed this in #203 to just overwrite with no-ops. Not sure of any negative implications. Input?

PR #164, opened 2018-05-06
Add `lolexFromGlobal` export

### LINKED ISSUE #146: Create fake timers based on passed in object

SimenB:
The first commit wraps the entire source inside of a factory function, and replaces global references with checks on the passed in `_global`. It then invokes that factory with `global || window` and exports the result from that call, to make the API stay the same for consumers. Highly recommended to view that diff without whitespace changes (https://github.com/sinonjs/lolex/pull/164/files?w=1).

The second commit just exports that factory function, and adds docs.

The name is horrible, suggestions more than welcome.

Fixes #146

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -273,6 +273,10 @@ Implements the `Date` object but using the clock to provide the correct time.
 
 Implements the `now` method of the [`Performance`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) object but using the clock to provide the correct time. Only available in environments that support the Performance object (browsers mostly).
 
+### `lolex.withGlobal`
+
+In order to support creating clocks based on separate or sandboxed environments (such as JSDOM), Lolex exports a factory method which takes single argument `global`, which it inspects to figure out what to mock and what features to support. When invoking this function with a global, you will get back an object with `timers`, `createClock` and `install` - same as the regular Lolex exports only based on the passed in global instead of the global environment.
+
 ## Running tests
 
 Lolex has a comprehensive test suite. If you're thinking of contributing bug
--- a/lolex.js
+++ b/lolex.js
@@ -2,800 +2,828 @@
 (function (global){
 "use strict";
 
-var userAgent = global.navigator && global.navigator.userAgent;
-var isRunningInIE = userAgent && userAgent.indexOf("MSIE ") > -1;
-var maxTimeout = Math.pow(2, 31) - 1; //see https://heycam.github.io/webidl/#abstract-opdef-converttoint
+function factory(_global) {
+    var userAgent = _global.navigator && _global.navigator.userAgent;
+    var isRunningInIE = userAgent && userAgent.indexOf("MSIE ") > -1;
+    var maxTimeout = Math.pow(2, 31) - 1; //see https://heycam.github.io/webidl/#abstract-opdef-converttoint
 
-// Make properties writable in IE, as per
-// http://www.adequatelygood.com/Replacing-setTimeout-Globally.html
-if (isRunningInIE) {
-    global.setTimeout = global.setTimeout;
-    global.clearTimeout = global.clearTimeout;
-    global.setInterval = global.setInterval;
-    global.clearInterval = global.clearInterval;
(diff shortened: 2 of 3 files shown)


lolex.js
```diff
@@ -2,800 +2,828 @@
 (function (global){
```
SimenB: should the diff from this file be excluded?

Readme.md
```diff
@@ -273,6 +273,10 @@ Implements the `Date` object but using the clock to provide the correct time.
 
 Implements the `now` method of the [`Performance`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) object but using the clock to provide the correct time. Only available in environments that support the Performance object (browsers mostly).
 
+### `lolex.lolexFromGlobal`
+
+In order to support creating clocks based on fake environments (such as JSDOM), Lolex exports a factory method which takes single argument `global`, which it inspects to figure out what to mock and what features to support. When invoking this function with a global, you will get back an object with `timers`, `createClock` and `install` - same as the regular Lolex exports only based on the apssed in global instead if the global environment.
```
Zirro: On the last line, "apssed" should presumably be "passed" and the "if" replaced with "of".

Could you perhaps refer to "separate environments" instead of "fake environments"? The latter description is not particularly accurate for jsdom.

SimenB: Good call! and thanks for the typo callouts 🙂

lolex.js
SimenB: ok, cool. removed the changes then 🙂

Readme.md
```diff
@@ -273,6 +273,10 @@ Implements the `Date` object but using the clock to provide the correct time.
 
 Implements the `now` method of the [`Performance`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) object but using the clock to provide the correct time. Only available in environments that support the Performance object (browsers mostly).
 
+### `lolex.lolexFromGlobal`
```
benjamingr: Any reason this has `lolex` in the name (rather than just `fromGlobal` or `withGlobal`?

SimenB: mostly as it kinda made sense in my head to do `const lolex = lolexFromGlobal(myGlobal)`. But I'm totally fine with another name. Any preferences?

SimenB:
I've tested with this PR in Jest now (https://github.com/facebook/jest/pull/5171) and it works pretty good!

benjamingr:
@fatso83 while I realise I can merge stuff - I think this change is big enough to ask for your opinion :)

Readme.md
fatso83: I'd like `withGlobal` or `fromGlobal`. Looks better IMHO `lolex.withGlobal(myJsDom)`

If people do static imports and think the contextless names are confusing they can name it whatever they like anyway:
`import {withGlobal as lolexFromGlobal} from 'lolex'`.

src/lolex-src.js
```diff
@@ -1,809 +1,825 @@
...
-    return clock;
-};
+    return {
+        timers: timers,
+        createClock: createClock,
+        install: install
+    };
+}
+
+var defaultImplementation = factory(global || window);
```
fatso83: Nice.

SimenB:
Renamed to `withGlobal` and squashed

fatso83:
Merged as da5505b (had to fix the commit msgs that didn't reflect the naming changes).

SimenB:
Oh, forgot to change the commit message when squashing. sorry about that!

PR #165, opened 2018-05-07
handle negative infinity timeout

SimenB:
See this issue in Jest, with links to spec: https://github.com/facebook/jest/issues/5960. Without changing the implementation of Lolex, Jest test suite breaks when attempting to use it

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -44,6 +44,18 @@ function withGlobal(_global) {
     var NativeDate = Date;
     var uniqueTimerId = 1;
 
+    function isNumberFinite(num) {
+        if (Number.isFinite) {
+            return Number.isFinite(num);
+        }
+
+        if (typeof num !== "number") {
+            return false;
+        }
+
+        return isFinite(num);
+    }
+
     /**
      * Parse strings like "01:10:00" (meaning 1 hour, 10 minutes, 0 seconds) into
      * number of milliseconds. This is used to support human-readable strings passed
@@ -201,6 +213,9 @@ function withGlobal(_global) {
         timer.type = timer.immediate ? "Immediate" : "Timeout";
 
         if (timer.hasOwnProperty("delay")) {
+            if (!isNumberFinite(timer.delay)) {
+                timer.delay = 0;
+            }
             timer.delay = timer.delay > maxTimeout ? 1 : timer.delay;
             timer.delay = Math.max(0, timer.delay);
         }
--- a/test/lolex-test.js
+++ b/test/lolex-test.js
@@ -280,6 +280,20 @@ describe("lolex", function () {
             this.clock.tick(61);
             assert.equals(callbackCalled, true);
         });
+        it("handles Infinity and negative Infinity correctly", function () {
+            var calls = [];
+            this.clock.setTimeout(function () {
+                calls.push("NaN");
+            }, NaN);
+            this.clock.setTimeout(function () {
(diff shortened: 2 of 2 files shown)


test/lolex-test.js
```diff
@@ -280,6 +280,20 @@ describe("lolex", function () {
...
+                calls.push("NaN");
+            }, NaN);
+            this.clock.setTimeout(function () {
+                calls.push("Infinity");
+            }, Number.POSITIVE_INFINITY);
+            this.clock.setTimeout(function () {
+                calls.push("-Infinity");
+            }, Number.NEGATIVE_INFINITY);
+            this.clock.runAll();
+            assert.equals(calls, ["NaN", "Infinity", "-Infinity"]);
```
SimenB: without the code change, `-Infinity` is before `Infinity` in the array

src/lolex-src.js
```diff
@@ -196,6 +208,9 @@ function addTimer(clock, timer) {
     timer.type = timer.immediate ? "Immediate" : "Timeout";
 
     if (timer.hasOwnProperty("delay")) {
+        if (!isNumberFinite(timer.delay)) {
```
benjamingr: So I checked - the spec actually says (if < 0 then = 0) but Infinity isn't explicitly treated - >Int32 numbers overflow and also get set to 0 but not explicitly. 

Node on the other hand sets the value to 1 (and not 0) https://github.com/nodejs/node/blob/master/lib/internal/timers.js#L19 done in https://github.com/nodejs/node/blob/master/lib/internal/timers.js#L55

benjamingr: https://github.com/nodejs/node/blob/master/lib/internal/timers.js#L55

SimenB: (I've never read much spec, so all questions are in good faith)

From the linked issue in Jest, I read it as point 8 in https://heycam.github.io/webidl/#abstract-opdef-converttoint (as it's a long, not `EnforceRange`), namely

>If x is NaN, +0, +∞, or −∞, then return +0.

benjamingr: Yeah, exactly.

SimenB: Oh, so we are in agreement. Cool! 😀 

Do you think it's a bug in Node?

benjamingr: @bnoordhois https://github.com/nodejs/node/commit/7fc835afe362ebd30a0dbec81d3360bd24525222 - 

Do you happen to remember why timers are set to 1 and not 0?

SimenB: Seeing as the tag failed, I'll be a bit cheeky and do @bnoordhuis

fatso83: So glad you guys are looking into the sweaty details of this. Delegation is :crown: Hope you add some comments to the implementation to make code archaeology easier for future maintainers :)

benjamingr: I've opened an issue over at core about the bahvior - I think it's fine to be honest, it's not what the timers spec says but it's what Chrome and Node do basically https://github.com/nodejs/node/issues/20596

SimenB: FWIW, I'm fine setting it to `1`. That's not observable from my test - the test is the order in which the callbacks gets invoked, not if the timeout is `0` or `1`.

fatso83:
2.4.1 out.

SimenB:
Awesome. Updated the PR in Jest. Getting closer!

Issue #167, opened 2018-05-08
Builds are failing on IE10 and others


fatso83:
After merging #160 the builds [started failing](https://travis-ci.org/sinonjs/lolex/jobs/367234969#L2533). We probably lack some checks to see if the `Performance` object exists at all.

SimenB:
Huh, we're actually hit by this in Jest in the PR moving to Lolex.

![image](https://user-images.githubusercontent.com/1404810/39768732-7659250c-52ea-11e8-8b01-1d3f2bdab1b8.png)

Probably just some faulty mocking. I can dig into it later this week if no one else gets to it 🙂

SimenB:
Dug into this (related to #162).

Using the fix in that PR is fine, but I'm unable to install it in JSDOM as it throws with `Cannot set property performance of #<Window> which has only a getter`. This does not match Chrome behavior, which allows you to set `window.performance = {}` just fine.

Not sure of the best way forward?

fatso83:
heads up: the fix in #162 is merged

PR #168, opened 2018-05-08
Access `Performance` via `_global`

SimenB:
Ref #167, but it's not enough. With this change, Jest fails with an internal JSDOM error...

image

Looks like the same error as the one from IE in #167, so fixing that will probably fix Jest as well


SimenB:
Might be a good integration test to pass a JSDOM global into `withGlobal` and asserting on some fake timers...

PR #169, opened 2018-05-08
add simple integration test with jsdom

SimenB:
Includes both #162 and #168, plus an integration test using JSDOM.

~~This doesn't really work though, as `window.performance` is `readonly` in JSDOM. This matches the spec (or, at least it matches the text on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Performance)), but it does not match the behavior of Chrome.~~

~~The second commit skips faking `performance`, and the test passes. I don't like that, though.~~

`window.performance` is `readonly`, so instead of using simple assignment, this PR uses `Object.defineProperty`.

### DIFF
--- a/integration-test/.eslintrc
+++ b/integration-test/.eslintrc
@@ -1,0 +1,12 @@
+{
+    "env": {
+        "mocha": true
+    },
+    "plugins": [
+        "mocha"
+    ],
+    "rules": {
+        "mocha/no-exclusive-tests": 2,
+        "max-nested-callbacks": 0
+    }
+}
--- a/integration-test/lolex-integration-test.js
+++ b/integration-test/lolex-integration-test.js
@@ -24,13 +24,14 @@ var sinon = require("sinon");
 var assert = referee.assert;
 
 describe("withGlobal", function () {
-    var jsdomGlobal, withGlobal;
+    var jsdomGlobal, withGlobal, timers;
 
     beforeEach(function () {
         var dom = new jsdom.JSDOM("", {runScripts: "dangerously" });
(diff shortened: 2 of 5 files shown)


SimenB:
I can look into having mochify skip the test same as node 4 if you want to keep the test.

mantoni:
See my change here for the `window.performace` issue: https://github.com/sinonjs/lolex/pull/162/files

SimenB:
@fatso83 any way of triggering the integration tests from a PR? I wonder if the last commit fixes the issue

fatso83:
@SimenB I don't think we have a way of triggering the full suite of Saucelabs tests on the PRs. It's been a long time, but I think the reason why we chose not to do it was that it would open us up to exposing the secret key (or something like that) by aggressors adding PRs that would read out environment variables. Or something like that ... @mantoni might remember it better 😊 

Anyway, I see you have your email listed on your profile, so I'll send you what you need so you can do it.

mantoni:
This is a [Travis security restriction](https://docs.travis-ci.com/user/pull-requests/#Security-Restrictions-when-testing-Pull-Requests), unfortunately.

I have a general question regarding this RP: Wouldn't switching to the latest Mochify with headless Chrome make a JSDom test superfluous? Or is JSDom something we want to have explicitly tested?

SimenB:
@fatso83 @benjamingr this now passes CI, and `npm run test-cloud` 🙂

mantoni:
👍 Looks good to me.

SimenB:
A release of this would be awesome 😀

Issue #159, opened 2018-04-03
Use a faster (more time efficient) data structure for clock.timers


akhilkedia:
**Summary** 
Right now, `clock.timers` in Lolex is an array, which is looped through completely everytime in `firstTimer()` or `LastTimer()`.
This is extremely inefficient, and perhaps other data structures such as a `heap` might be much faster.

**Background Details**
Lolex is great for running simulations - but in one of my recent projects, I attempted to use Lolex to run a code over a simulated period of 1 week - the code made a LOT of `setTimeouts`, and these were all handled by Lolex.

Profiling the code showed the majority of my simulation runtime was spent in Lolex's `firstTimer()` function, because it kept looping through a long array of timers everytime.

Changing `clock.timers` to a simple heap fixed this performance bottleneck.

Since Lolex is targeted primarily towards running simulations/tests, being able to run them faster will be great.

benjamingr:
Actually, timers is a hashmap (well object) and we iterate the hashmap for the first one, if it were an array and it was kept sorted then inserts would be pretty fast and firstTimer would be similarly fast.

akhilkedia:
My bad, yes timers is a Hashmap (object.)

Even if you change timers to a sorted array, while first/last and removing a timer could be fast, insertion into a sorted array would still take `O(n)` time (n is length of timers array) as the other elements in the array would need to be shifted.

Hence I suggest using any priority queue structure, such as a heap.

benjamingr:
@akhilkedia any interest in working on this? I can point you to the right places in the code if you'd like :) 

If not I can take a stab next month probably.

Alexsey:
@benjamingr do you mean that this https://github.com/sinonjs/lolex/blob/master/src/lolex-src.js#L345 is an error and it must be `{}` and not `[]`?

And we may stop using ugly `hasOwnProperty` if `timers` would be created with `Object.create(null)`

And about `delete` - it is [still slow](https://www.nearform.com/blog/node-js-is-getting-a-new-v8-with-turbofan/) so, probably, better would be to use `= undefined`. Especially, if we already are trying to protect ourselves from `Object`'s prototype manipulations (usage of `hasOwnProperty`) then `delete` would not save us. But `= undefined` would

benjamingr:
> @benjamingr do you mean that this https://github.com/sinonjs/lolex/blob/master/src/lolex-src.js#L345 is an error and it must be {} and not []?

Yeah, though given how JS works it doesn't really matter.

[...]

PR #173, opened 2018-05-12
fix: change clock.timers from array to prototype-less object

Alexsey:

Fix issue #159 - Use a faster (more time efficient) data structure for clock.timers

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -231,7 +231,7 @@ function withGlobal(_global) {
         }
 
         if (!clock.timers) {
-            clock.timers = {};
+            clock.timers = Object.create(null);
         }
 
         timer.id = uniqueTimerId++;
@@ -294,13 +294,11 @@ function withGlobal(_global) {
         var timer = null;
         var id, isInRange;
 
-        for (id in timers) {
-            if (timers.hasOwnProperty(id)) {
-                isInRange = inRange(from, to, timers[id]);
+        for (id in timers) { // eslint-disable-line guard-for-in
+            isInRange = inRange(from, to, timers[id]);
 
-                if (isInRange && (!timer || compareTimers(timer, timers[id]) === 1)) {
-                    timer = timers[id];
-                }
+            if (isInRange && (!timer || compareTimers(timer, timers[id]) === 1)) {
+                timer = timers[id];
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/lolex-src.js
```diff
@@ -231,7 +231,7 @@ function withGlobal(_global) {
         }
 
         if (!clock.timers) {
-            clock.timers = {};
+            clock.timers = Object.create(null);
```
benjamingr: Can you also change the other definition of `clock.timers`?

benjamingr:
This means lolex can't support ES3 environments anymore (ES3 didn't have objects with null prototypes) - are those currently supported?

For the record - IE8 is the latest ES3 environment.

benjamingr:
Also, it does not fix #159 - since you're still using an object rather than a PriorityQueue so everything is iterated every time a timer is looked for anyway.

Alexsey:
About ES3 - Sinon says in its docs that [it's ES5.1](http://sinonjs.org/releases/v5.0.7/). There is nothing about supported ES version in lolex Readme so I decide that it is also targeting ES5.1

About priority queue - hashmap (object) is giving us [average O(1) on search, O(1) on insert and O(1) on delete](https://en.wikipedia.org/wiki/Hash_table) while all heaps would require [O(log n) at least on delete](https://en.wikipedia.org/wiki/Heap_(data_structure)). But if you think it could be faster - than ok, this is not closing #159 - it's just a bugfix

benjamingr:
I don't think this change is a significant improvement to be honest since it doesn't actually make lolex faster. I don't feel strongly about this though.

fatso83:
Yeah, I would guess you'd need to be in a pretty tight loop to ever notice the lack of guards in an actual scenario, but for what I know this might well be the case resulting in 5% improvement in some special case :man_shrugging: 

I don't think this is a significant improvement either, but if it's less code doing the same work in a clearer manner then it's usually a minor refactoring worth shoving in.

benjamingr:
> Yeah, I would guess you'd need to be in a pretty tight loop to ever notice the lack of guards in an actual scenario,

`for...in` actually performs better with the `hasOwnProperty` check in V8 since V8 explicitly knows that construct. I'm not really up to date on what other browsers do though.

Alexsey:
@fatso83 I can remove all `Object.create(null)` and leave only fix of initialization of `clock.timers` with array while it should be an object. In this case, would you accept this PR? Actually, that is a bug that doesn't affect logic but confuses anyone who reads the code

fatso83:
Sure, I failed to see the fix in line 365.

benjamingr:
@Alexsey we'd love to have more contributions from you to be clear - I hope the fact some of the changes here didn't make it wouldn't discourage you from working on more improvements :)

Alexsey:
@benjamingr I understand that not every PR should be accepted - it's absolutely ok as long as there is some feedback. So here everything is ok :)

PR #178, opened 2018-05-15
 Access Date on `_global`

SimenB:
This was discovered by @mjesun when trying to apply a version of Jest with Lolex at Facebook.

I'm not sure what the solution should be. Our idea was to just assign the `prototype`, but it is already copied

### DIFF
--- a/integration-test/lolex-integration-test.js
+++ b/integration-test/lolex-integration-test.js
@@ -48,4 +48,14 @@ describe("withGlobal", function () {
 
         clock.uninstall();
     });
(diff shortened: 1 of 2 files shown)


SimenB:
Found the fix! It was me in #164 not being thorough enough

mjesun:
LGTM to me! But that's not Jest, I can't merge 😄

I can apply the patch internally, though, and trigger a second run.

benjamingr:
Thanks!

PR #179, opened 2018-05-16
Clear microtick jobs and set now to 0 in reset

### LINKED ISSUE #161: Is method reset() public?
There is a [reset](https://github.com/sinonjs/lolex/blob/master/src/lolex-src.js#L692) method in the code but it is not in the docs (readme)

1. May I rely on it?
2. May I add to it `clock.now = 0`?

I can make a PR with it to docs

SimenB:

See #161 for justification. Fixes #161

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -722,6 +722,8 @@ function withGlobal(_global) {
 
         clock.reset = function reset() {
             clock.timers = {};
+            clock.jobs = [];
+            clock.now = 0;
         };
 
         clock.setSystemTime = function setSystemTime(systemTime) {
--- a/Readme.md
+++ b/Readme.md
@@ -225,6 +225,10 @@ callbacks.
 
 Advances the clock to the the moment of the first scheduled timer, firing it.
 
+### `clock.reset()`
+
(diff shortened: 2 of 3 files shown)


src/lolex-src.js
```diff
@@ -722,6 +722,8 @@ function withGlobal(_global) {
 
         clock.reset = function reset() {
             clock.timers = {};
```
benjamingr: Technically this can `delete clock.timers` and `delete clock.jobs`  to actually "reset" it since they are added lazily to the clock - though no strong felings.

Are you using `.reset` for jest?

benjamingr: Then I think we can do this - it seems reasonable and since lolex is mostly used by bigger consumers rather than independently I think that a big consumer (jest) wanting the feature makes it worthwhile exposing - especially since it's already implemented and probably used but doesn't work well.

I think the current approach you're using is fine - I was just pointing it out.

fatso83:
Benjamin, you have already been added to the Sinon organisation on NPM. Try it out yourself :-) Read RELEASE.md.

benjamingr:
Going to give it a shot!

Issue #171, opened 2018-05-11
Performance is not defined


daisyxu0924:
Hello, I'm getting
 ``` ReferenceError: Performance is not defined
      at createClock (node_modules/lolex/src/lolex-src.js:736:36)
      at Object.install (node_modules/lolex/src/lolex-src.js:791:21)
      at Sandbox.exports.useFakeTimers (node_modules/sinon/lib/sinon/util/fake_timers.js:26:21)
```
with the new release 2.4.1. 

```
├─┬ sinon@4.5.0
│ ├─┬ @sinonjs/formatio@2.0.0
│ │ └── samsam@1.3.0
│ ├── lodash.get@4.4.2
│ ├── lolex@2.4.1
```
When used lolex@2.3.2 it was working fine.

Thanks!

**Comment by benjamingr** [comment]
Hey, apologies! @daisyxu0924 

Can you please share what browser you're seeing this in?

**Comment by daisyxu0924** [comment]
Hello @benjamingr sorry about the late reply. my useragent is 
```Mozilla/5.0 (darwin) AppleWebKit/537.36 (KHTML, like Gecko) jsdom/11.7.0```
I've updated lolex to 2.5.0 global.Performance is still undefined. Thanks!

**Comment by phoebejhhong** [comment]
Hey I'm working on the same project with @daisyxu0924

The code that invokes this problem is just
```js
sinon.useFakeTimers(Date.now())
```

as part of our test suite along with mocha, enzyme and webpack.
I've found the commit that caused problem for us (https://github.com/sinonjs/lolex/commit/469b547334aeee8731581a24b8e42fb23d2aea43) and when we comment that block of code out, everything is fine. After version 2.4.2, the error message was updated to`TypeError: Cannot read property 'prototype' of undefined` because `global.Performance` is undefined.

dwoowb:
@daisyxu0924 I'm running into what looks like the same issue. Could you elaborate a bit on how jsdom was related to this issue for you? Thank you!

daisyxu0924:
Hi @dwoowb we are using enzyme + jsdom and running tests in node env. So what we did was overriding the node global env with jsom (which is not recommended by jsdom btw). So when lolex is executing global.Performance it's actually reading the node env. If yours is the same case, have a look at this: https://github.com/airbnb/enzyme/blob/master/docs/guides/jsdom.md

panzerdp:
[...]

PR #181, opened 2018-05-16
Check Performance exists before using it

benjamingr:

Fixes https://github.com/sinonjs/lolex/issues/171#issuecomment-389487264 since apparently some environments expose `global.performance` but not `global.Performance`. 

#### Solution  - optional

Guard checking performance with `global.Performance`.

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -31,7 +31,9 @@ function withGlobal(_global) {
     var addTimerReturnsObject = typeof timeoutResult === "object";
     var hrtimePresent = (_global.process && typeof _global.process.hrtime === "function");
     var nextTickPresent = (_global.process && typeof _global.process.nextTick === "function");
-    var performancePresent = (_global.performance && typeof _global.performance.now === "function");
+    var performancePresent = (_global.performance &&
+        _global.Performance &&
+        typeof _global.performance.now === "function");
     var requestAnimationFramePresent = (
         _global.requestAnimationFrame && typeof _global.requestAnimationFrame === "function"
     );
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -31,9 +31,8 @@ function withGlobal(_global) {
     var addTimerReturnsObject = typeof timeoutResult === "object";
     var hrtimePresent = (_global.process && typeof _global.process.hrtime === "function");
     var nextTickPresent = (_global.process && typeof _global.process.nextTick === "function");
-    var performancePresent = (_global.performance &&
-        _global.Performance &&
-        typeof _global.performance.now === "function");
+    var performancePresent = (_global.performance && typeof _global.performance.now === "function");
+    var performanceConstructorExists = (_global.Performance && typeof _global.Performance === "function");
     var requestAnimationFramePresent = (
         _global.requestAnimationFrame && typeof _global.requestAnimationFrame === "function"
     );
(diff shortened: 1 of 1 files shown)


SimenB:
This means environments with `performance.now` but no `Performance` will _not_ be mocked, right?

Would it be better to guard around specifically https://github.com/sinonjs/lolex/blob/f7d11449c955556652d662b463f9e42903d41903/src/lolex-src.js#L751-L759 but leave `performance.now` to be mocked?

benjamingr:
@SimenB Looking at https://developer.mozilla.org/en-US/docs/Web/API/Performance and https://developer.mozilla.org/en-US/docs/Web/API/Performance/now I don't really understand what environments do one but not the other.

benjamingr:
BTW - If you feel strongly about it I'll gladly edit the PR. I just want a better understanding of where this problem bites people.

Another option is to get `Performance` by `performance.__proto__` (performance doesn't exist in IE which doesn't have __proto__ - we can also do Object.getPrototypeOf).

SimenB:
Doesn't impact me, just throwing out suggestions 🙂

benjamingr:
Ok, let's wait to see if the user who complained shows up to provide an environment we can produce a failing case for then :)

Kyr:
Please, not only check is Performance or performance present, this [fragment]( https://github.com/sinonjs/lolex/blame/f7d11449c955556652d662b463f9e42903d41903/src/lolex-src.js#L751-L759) cause serious problem on node, 8, 10, whatever due to Performance does not present, and more over:
```
> perf_hooks.performance.__proto__
Performance {}
> perf_hooks.performance.prototype
undefined
```
so any sinon's fake timers does not work

benjamingr:
@Kyr thanks for stepping in and contributing. Note that `Performance` is the performance prototype - which is the same as `performance.__proto__`

So generally since `performance instanceof Performance` that means `performance.__proto__ === Performance.prototype`. Note the uppercase `P`

benjamingr:
Also, in Node 8 performance isn't exported a global which would mean the entire piece of code wouldn't get hit at all initially. If you have code reproducing the issue I'd be very interested in seeing it.

Thanks again.

Kyr:
Thanks for explanation @benjamingr , but may I apoint you to something another
image

node v8.11.2

benjamingr:
@Kyr thanks a lot. Can you tell us what environment this is running in? (Node?)

Can you please share:

 - What version is running (Node 8.x?)
 - What's the value of `_global`? 
 - What's the value of `_global.performance`?
 - Is this a JSDOM test (or something similar to a JSDOM test?)

Kyr:
@benjamingr I apologize for being misled, I completely forgot about the years made that custom jsdom setup which broke everything.

After removing already unnecessary patches, everything started to work fine.

benjamingr:
> After removing already unnecessary patches, everything started to work fine.

I would still like to make sure that we don't break user deployments. I'm a firm believer in lolex not breaking peoples' code and I'll do better in the future.

Thanks for the report - that's helpful. 

@SimenB given the above - what do you think we should do?

benjamingr:
Looks like this is an "overriding globals" issue https://github.com/jsdom/jsdom/issues/2231 - given that and not a misconfiguration I'm going to go ahead and go with @SimenB's suggested approach

benjamingr:
Updated, PTAL

src/lolex-src.js
```diff
@@ -750,13 +751,15 @@ function withGlobal(_global) {
 
             var proto = _global.Performance.prototype;
```
SimenB: Missing a guard here

SimenB: Can probably reuse `proto` on line 756

benjamingr: Yeah, makes sense

Issue #183, opened 2018-05-21
reset should set "now" to the same value clock was installed


Alexsey:
* Lolex version : 2.6.0
* Environment   : NodeJS

**What did you expect to happen?**
```js
const lolex = require('lolex');
const clock = lolex.install({now: 10000})
clock.reset();
console.log(clock.now) // 10000
```

**What actually happens**
```js
const lolex = require('lolex');
const clock = lolex.install({now: 10000})
clock.reset();
console.log(clock.now) // 0
```

The current behavior of `reset` is killing the sense of "now" parameter because it's getting lost after the first `reset` call

benjamingr:
On it

benjamingr:
Thanks for the report! Can you please verify the fix in https://github.com/sinonjs/lolex/pull/184 works for you?

Alexsey:
For now, I have only one test case but this fix fork for it. Nice!

benjamingr:
#184 is merged so I'm going to go ahead and close this. Feel free to reopen if you think there is still pending work to be done here.

PR #184, opened 2018-05-21
Hotfix/clock reset time

benjamingr:

Fix issue https://github.com/sinonjs/lolex/issues/183 by setting the time to the same value as it is set initially.


SimenB:
Ah, nice. Way better!

PR #186, opened 2018-05-21
fix reset docs

Alexsey:
Reflect changes of https://github.com/sinonjs/lolex/pull/184


benjamingr:
Thanks

Issue #191, opened 2018-05-31
ref(), & unref() return value


icebob:
> We understand you have a problem and are in a hurry, but please provide us with some info to make it much more likely for your issue to be understood, worked on and resolved quickly.

* Lolex version : 2.6.0
* Environment   : NodeJS v8.11

**What did you expect to happen?**
According to NodeJS documentation the `ref()` and `unref()` return the timer.
https://nodejs.org/dist/latest-v8.x/docs/api/timers.html#timers_timeout_ref

**What actually happens**
But lolex returns undefined.

**How to reproduce**
```js
const lolex = require("lolex");
const clock = lolex.install();

const timer = setTimeout(() => {}, 5000).unref();
console.log(timer);
// Prints: undefined
```
If you comment out the `lolex.install()` line, it prints the NodeJS Timeout object.

Thank you!

fatso83:
Grep the lolex code for `unref` and see what it does (or look at the code, it's just a couple of hundred lines in one file). I haven't encountered that before. Maybe we just define the function but no implementation. Wouldn't know, but I do know we accept PRs ;)

icebob:
yes, I checked the code, only need to change ~2 lines. I'm making a PR.

PR #193, opened 2018-06-01
Fixes ref() & unref() return value on NodeJS

icebob:
Fix issue #191 

According to NodeJS documentation the ref() and unref() return the timer.
https://nodejs.org/dist/latest-v8.x/docs/api/timers.html#timers_timeout_ref

This PR fixes it in lolex. Relevant tests are added.

After fix it prints the correct timer:
```js
const lolex = require("./src/lolex-src");

const timer = setTimeout(() => {}, 5000).unref();
console.log(timer);
// Prints: { id: 1, ref: [Function: ref], unref: [Function: unref] }
```

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -245,11 +245,12 @@ function withGlobal(_global) {
         clock.timers[timer.id] = timer;
 
         if (addTimerReturnsObject) {
-            return {
+            var res = {
                 id: timer.id,
-                ref: NOOP,
-                unref: NOOP
+                ref: function () { return res; },
+                unref: function () { return res; }
             };
+            return res;
         }
 
         return timer.id;
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -248,7 +248,8 @@ function withGlobal(_global) {
             var res = {
                 id: timer.id,
                 ref: function () { return res; },
-                unref: function () { return res; }
+                unref: function () { return res; },
+                refresh: function () { return res; }
             };
             return res;
         }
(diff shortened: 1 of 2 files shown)


src/lolex-src.js
```diff
@@ -245,11 +245,12 @@ function withGlobal(_global) {
         clock.timers[timer.id] = timer;
 
         if (addTimerReturnsObject) {
-            return {
+            var res = {
                 id: timer.id,
-                ref: NOOP,
-                unref: NOOP
+                ref: function () { return res; },
+                unref: function () { return res; }
```
benjamingr: Can you add `.refresh` here?

icebob: What is the `.refresh`?

benjamingr: It's a new API Node recently added in https://github.com/nodejs/node/pull/20298 that landed in 10.2.0

benjamingr: If you don't feel like it I'll land and can release when I get back from the Node summit/JSConf next week.

Just saying I would appreciate it if you did it while touching that code - this will land regardless - no pressure.

Already very happy that you decided to contribute to the project :)

icebob: Ohh, I've not known about it. But I'm adding it.

icebob: Done.

fatso83: I think adding [the whole logic](https://github.com/nodejs/node/pull/20298/files#diff-5e13dd343d32f1dc04d5b21db8979471R77) is a bit much in this (essentially unrelated PR), but adding a `NOOP` should be sufficient. Someone else can supply a full PR.

PR #195, opened 2018-07-02
Fix performance replacement on iOS 9.3

### LINKED ISSUE #190: Safari 9 compatibility is broken
* Lolex version : 2.7
* Environment   : Safari 9

#169 assigns props in a way that's not compatible with Safari 9. That broke tests in Sinon (sinonjs/sinon#1814), which shows that we missed a browser in our browser suite (added yesterday). We should aim to support the same browsers as Sinon, given that it's the main target to support.

I expected the `test-cloud` script to pass

**What actually happens**
It fails

**How to reproduce**
Set the Sauce Labs credentials (known to Sinon maintainers) as environment variables.
`npm run test-cloud`

mantoni:

While trying to run a test suite with fake timer in iOS 9.3, I discovered an issue with lolex: `window.performance` cannot be replaced using `Object.defineProperty` because its descriptor has `configurable: false`. However, assigning a new value works.

Also happens to fix #190.

#### Solution  - optional

I first tried checking on the descriptor `configurable` property which fixed the issue. It turned out that simply replacing the property works as well, but I recalled that `Object.setProperty` was introduced in #169 by @SimenB to fix it in JSDom. However, all tests pass and I'd like to have confirmation by a JSDom expert that this change doesn't re-introduce the issue.

#### Verifying the problem

I can reproduce the issue with the lolex test suite:

```bash
$ node_modules/.bin/mochify --consolify test.html
$ http-server .
```

and then open `http://localhost:8080/test.html` in the iOS 9.3 emulator.

With this fix applied, the test suite passes again.

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -405,10 +405,7 @@ function withGlobal(_global) {
             } else if (method === "nextTick" && target.process) {
                 target.process.nextTick = clock[installedNextTick];
             } else if (method === "performance") {
-                Object.defineProperty(target, method, {
-                    writeable: false,
-                    value: clock["_" + method]
-                });
+                target[method] = clock["_" + method];
             } else {
                 if (target[method] && target[method].hadOwnProperty) {
                     target[method] = clock["_" + method];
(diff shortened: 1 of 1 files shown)


SimenB:
Since the integration test added passes, I'm happy! 🙂

benjamingr:
I completely forgot about this issue! Sorry! LGTM from my end if it works with JSDom as @SimenB noted.

Will defer to @fatso83 for merging

mroderick:
This has been published as `lolex@2.7.1

Issue #198, opened 2018-08-15
Error in browser: Uncaught ReferenceError: global is not defined


budarin:
* Lolex version : 2.7.1
* Environment   : Chrome 68.0.3440.106, Windows
* Other libraries you are using: no

**What did you expect to happen?**
expet to have window.lolex to be accesible

**What actually happens**
error at:
`var defaultImplementation = withGlobal(global || window);`

**How to reproduce**
to include tag into page's head
`<script src="https://cdn.jsdelivr.net/npm/lolex@2.7.1/src/lolex-src.min.js" crossorigin></script>`
even if to patch this
`global = window`

next error occures at 
`exports`

benjamingr:
Checking.

benjamingr:
PR at https://github.com/sinonjs/lolex/pull/199 - sorry for this! Thanks for the report.

benjamingr:
You need https://cdn.jsdelivr.net/npm/lolex@2.7.1/lolex.min.js and not https://cdn.jsdelivr.net/npm/lolex@2.7.1/src/lolex-src.min.js

budarin:
thanks, it solves the problem!

Issue #197, opened 2018-07-12
TypeError: Cannot set property performance of #<Object> which has only a getter


daisyxu0924:
> We understand you have a problem and are in a hurry, but please provide us with some info to make it much more likely for your issue to be understood, worked on and resolved quickly.

* Lolex version : 2.7.1
* Environment   : jsdom window
* Example URL   : 
* Other libraries you are using:

**What did you expect to happen?**

**What actually happens**
Hello, we use sinon's useFakeTimers after copying jsdom window to global following this method https://github.com/airbnb/enzyme/blob/master/docs/guides/jsdom.md

after https://github.com/sinonjs/lolex/commit/79f9431a7278509ebffbe107484f521f76b3f665 this commit, it's complaining:

```
TypeError: Cannot set property performance of #<Object> which has only a getter
      at hijackMethod (node_modules/lolex/src/lolex-src.js:445:28)
      at Object.install (node_modules/lolex/src/lolex-src.js:844:17)
      at exports.useFakeTimers (node_modules/sinon/lib/sinon/util/fake_timers.js:26:21) 
```

and 

```
TypeError: Cannot assign to read only property 'performance' of object '#<Object>'
      at uninstall (node_modules/lolex/src/lolex-src.js:408:32)
      at Object.clock.uninstall (node_modules/lolex/src/lolex-src.js:817:20)
```

 To be honest I'm not sure if I should be asking JSDOM to fix the window object or should lolex's code handle this scenario, anyway I'll just post this issue here first. 

Thanks!
**How to reproduce**
> Describe *with code* how to reproduce the faulty behaviour
> or link to code on JSBin or similar

<details>
    <summary>Really long code sample or stacktrace</summary>

    If you need to provide a dump of a stack trace or
    other lengthy material, such as 80 lines of example code,
    please stuff it in a `<details>` tag such as this
    to make the issue more readable. Thanks.
</details>

philostler:
Just bumped into the same issue here. @daisyxu0924 can you please cross reference that issue JSDOM if you've created one

fatso83:
Seems solved by #199

shockey:
I'm seeing a very similar error with `lolex@3.0.0`:

```
     TypeError: Cannot set property performance of #<Object> which has only a getter
      at hijackMethod (node_modules/lolex/src/lolex-src.js:445:28)
      at Object.install (node_modules/lolex/src/lolex-src.js:877:17)
[...]

PR #199, opened 2018-08-15
fix: global related bugs

benjamingr:

Fixes https://github.com/sinonjs/lolex/issues/198 and https://github.com/sinonjs/lolex/issues/197

Needs review (cc @fatso83 )
<!--
#### Solution  - optional

 - guard variable access with `typeof`
 - guard assignment with fallback to the old jsdom friendly method

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -441,7 +441,14 @@ function withGlobal(_global) {
             var date = mirrorDateProperties(clock[method], target[method]);
             target[method] = date;
         } else if (method === "performance") {
-            target[method] = clock[method];
+            try {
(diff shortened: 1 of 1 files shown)


mAAdhaTTah:
This won't solve the error with `exports` being undefined in the browser.

benjamingr:
@mAAdhaTTah actually, I'm not sure how you got that issue - I just tested and I wasn't able to reproduce it when testing carefully - browserify is supposed to take care of this.

mAAdhaTTah:
Oh @budarin mentioned that in the other issue, but if it's resolved, then no worries!

Issue #201, opened 2018-09-04
Sinon Fake Timers: setInterval() does not work as expected


fatso83:
_From @ggsbv on August 28, 2018 16:24_

Sinon's implementation of setInterval() does not work when a string is supplied as the interval.

1. Open Node REPL in the terminal.
2. Do the following and observe the output
```
const sinon = require('sinon')
const clock = sinon.useFakeTimers()
const blah = () => console.log(Date.now())

// passing a string
setInterval(blah, '100')

clock.tick(100)
clock.tick(100)
clock.tick(100)

// passing an int
setInterval(blah, 100)

clock.tick(100)
clock.tick(100)
clock.tick(100)
```

**Expected behavior**
You would expect setInterval to return the intervalID value at each invocation, incrementing by 100 each time. You would expect the same behaviour in both scenarios, regardless of whether the interval value is a string or an integer.

**Screenshots**
![screen shot 2018-08-28 at 17 22 04](https://user-images.githubusercontent.com/20184132/44736358-ed958800-aae6-11e8-9927-ef6f4d909c84.png)

_Copied from original issue: sinonjs/sinon#1888_

fatso83:
_From @mgred on August 30, 2018 14:21_

@ggsbv 
Just a guess... I think this is not related to sinon, but to [lolex](https://github.com/sinonjs/lolex) which is the package in use here. 

The `setInterval` Implentation returns an invocation of [`addTimer`](https://github.com/sinonjs/lolex/blob/master/src/lolex-src.js#L212), which checks if the given delay (timeout) is a finite number `Number.isFinite()`. That returns `false` and delay is set to 0 in case you put in a string value.

I hope that helps in any way...

fatso83:
@ggsbv The entire fix to this issue is just to parse the args before adding the timer. Check out [this line](https://github.com/sinonjs/lolex/blob/master/src/lolex-src.js#L549). Want to submit [your first PR](https://github.com/sinonjs/lolex#running-tests) to lolex? <3<3<3

fatso83:
Fixed by #202

PR #202, opened 2018-09-04
fix(setInterval): parse `timeout` arg to integer

mgred:
#### Purpose (TL;DR)

Fix issue #201 by always parsing the `timeout` argument of `setInterval` to an integer value, so that strings are possible.

#### Background

See Description in #201 

#### Solution

As suggested by @fatso83, the argument is now parsed to an integer value (`parseInt()`).
There is also a test proving that a string input works.

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -544,6 +544,7 @@ function withGlobal(_global) {
             });
         };
         clock.setInterval = function setInterval(func, timeout) {
+            timeout = parseInt(timeout);
             return addTimer(clock, {
                 func: func,
                 args: Array.prototype.slice.call(arguments, 2),
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -544,7 +544,7 @@ function withGlobal(_global) {
             });
         };
         clock.setInterval = function setInterval(func, timeout) {
-            timeout = parseInt(timeout);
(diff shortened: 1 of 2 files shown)


fatso83:
Thanks!

mroderick:
Thank you 👍

mgred:
You're welcome guys and Thank You for the awesome sinon projects!!!!

src/lolex-src.js
```diff
@@ -544,6 +544,7 @@ function withGlobal(_global) {
             });
         };
         clock.setInterval = function setInterval(func, timeout) {
+            timeout = parseInt(timeout, 10);
```
benjamingr: Is the radix correct?

Looking at the spec - I think `Number` is more correct here. It's really a tiny nit, but if someone passes  `0x100` to setInterval or setTimeout the correct timeout according to the timer spec is 255 and not 0 (which this produces).

Not a big deal by any measure.

mgred: True, the parse function assumes to get a decimal number (10). I think we could leave this off to let the function decide which radix to use depending on the numerical system.

What do you mean with
> I think `Number` is more correct here

Maybe to use the `Number.parseInt()` function? This should be the same as the global one afaik.

Issue #200, opened 2018-08-17
TypeError: 'mark' called on an object that does not implement interface Performance


sanchit-kialo:
* Lolex version: 2.7.1
* Environment: Firefox 62.0b17 - macOS 10.13.6
* Other libraries you are using: ReactDOM (development)

**What did you expect to happen?**
No Error to be thrown

**What actually happens**
A TypeError was thrown
`TypeError: 'mark' called on an object that does not implement interface Performance`

**How to reproduce**
`lolex.install();`
This happens when using the default `lolex.install();`

When subsequently calling `performance.mark()`, the TypeError is thrown (in my case, `performance.mark` is called by ReactDOM.development)

A workaround for this issue that I have found is to use this instead:
`lolex.install({
            toFake: [
                'setTimeout',
                'clearTimeout',
                'setImmediate',
                'clearImmediate',
                'setInterval',
                'clearInterval',
                'Date',
                'nextTick',
                'hrtime',
            ],
        });`

Based on my preliminary research, it looks like this issue may have been introduced by #160 in response to issue #136 
I am unsure if this method of recreating the performance object on the clock is carrying over the underlying interface type.

fatso83:
Veriifed by adding this test and running `npm run test-headless`

```javascript
            it("should not throw an error on calling performance.mark", function() {
                this.clock = lolex.install();
                assert.isFunction(global.performance.mark);
                refute.exception(function() {
                    global.performance.mark("a name");
                });
            });
```

PR #203, opened 2018-09-05
Fix issue 200: performance.mark triggers TypeError

fatso83:
Fixes issue #200 by overriding methods on window.performance

Could not see any reason why to keep the initial implementation delegated to the original
object, so overrode them with no-ops. Input on why this might not be the right way
to deal with this?

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -754,7 +754,7 @@ function withGlobal(_global) {
                     .getOwnPropertyNames(proto)
                     .forEach(function (name) {
                         if (Object.getOwnPropertyDescriptor(proto, name).writable) {
-                            clock.performance[name] = proto[name];
+                            clock.performance[name] = NOOP;
                         }
                     });
             }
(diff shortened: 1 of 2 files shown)


fatso83:
The regression test only (originally) failed on `npm run test-headless`, so use that when verifying the regression test.

fatso83:
@SimenB I just realized Ben would be unavailable for the next 6 weeks or so (based on some message I saw). Would you have a chance of reviewing this? I can't really see a reason to keep the other `performance` methods, as it is stands, but then again; I never really saw the usecase for overriding `performance.now` to begin with :blush:

fatso83:
I think that's what [the first commit](https://github.com/sinonjs/lolex/pull/203/commits/0c66e12782fe769d3dd52bf6f0331239ed39f338#diff-83f5b888d9039d7250c8a6c73e051ad5R1942) is doing? Or do I fail to understand you?

fatso83:
Hmm ... I was a bit quick to merge that, as I forgot I asked for input from Max, but in case there are issues I can patch it up later :-/ Publishing to NPM asap.

fatso83:
Damn, the cloud tests failed after publish.

fatso83:
#204 should fix the stuff that failed in the Sauce Labs tests

PR #204, opened 2018-09-05
Fix cloud tests for IE and Safari :hankey:

fatso83:
The changes introduced for 2.7.3 did not break anything, but it introduced a new test that showed
bugs in the existing implementation by actually running `performance.mark()`.

Also, trying to run performance.mark failed on Safari 9, as it has `performance.now`, but
not `performance.mark`. We therefore needed an additional guard before those tests.

We didn't see the fail until merging to master due to pull requests not running the Sauce Labs
tests, and me being too forgetful and impatient to wait for the tests 😀

The changes can be split into two

- changes in the test code to avoid running a test in Safari 9
- changes in the library that deals with IE 10 and IE 11

The prototype for `Performance` in Internet Explorer is not a function like
everywhere else, so we needed to expand the test for object types.


fatso83:
Btw, this has been tested extensively in all the browsers using `npm run test-cloud`, which is why the fix took so long.
one bug down, another one arises!

PR #208, opened 2018-09-17
fix: runToLast should respect current time

SimenB:
Without the code change, the added test makes node hangs forever until it runs out of memory.

`clock.tick` expect the argument to be how many millis it should tick. Currently, we pass it the timestamp of when it should be done ticking.

https://github.com/sinonjs/lolex/blob/1278eaade7da311d203fc6c33f67f85c828a09f2/src/lolex-src.js#L723

I don't understand how no-one hit this bug for 2.5 years... It's been that way since the original implementation: #50.

#### Solution

Subtract the current time on the clock before calling `tick`. The first thing tick does is add it back, so it knows when to stop.

https://github.com/sinonjs/lolex/blob/1278eaade7da311d203fc6c33f67f85c828a09f2/src/lolex-src.js#L603

We might want a `tickTo` method? Implementation would be `this.tick(to - clock.now)`, so maybe not valuable.

### DIFF
--- a/test/lolex-test.js
+++ b/test/lolex-test.js
@@ -1225,21 +1225,17 @@ describe("lolex", function () {
         it("should support clocks with start time", function () {
             var startingPoint = new Date("2018-01-01").getTime();
             this.clock = lolex.createClock(startingPoint);
-            var runOrder = [];
             var that = this;
+            var invocations = 0;
 
             this.clock.setTimeout(function cb() {
-                runOrder.push("mock1");
-                that.clock.setTimeout(cb, 49);
+                invocations++;
+                that.clock.setTimeout(cb, 50);
             }, 50);
-
-            this.clock.setTimeout(function cb() {
-                runOrder.push("mock2");
-            }, 100);
 
             this.clock.runToLast();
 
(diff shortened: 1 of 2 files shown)


SimenB:
Also, I would have expected the timers to bail after the max number of iterations were reached, not keep scheduling new ones until we run out of memory

test/lolex-test.js
```diff
@@ -1222,6 +1222,22 @@ describe("lolex", function () {
             assert.isTrue(spy.called);
         });
 
+        it("should support clocks with start time", function () {
+            var startingPoint = new Date("2018-01-01").getTime();
+            this.clock = lolex.createClock(startingPoint);
```
SimenB: This is a stupidly high number (`1514764800000`), but if you do e.g. `200` instead the test will fail with `invactions` being equal to `5` instead of hanging forever

SimenB:
Simplified the test per my comment 🙂 

Would love a release of this, as I'm trying to pick up Lolex in Jest again, and this is blocking

fatso83:
Out as 2.7.5

PR #209, opened 2018-09-18
fix: reset high resolution timer on clock.reset

### LINKED ISSUE #206: reset() does not reset high-resolution timer
* Lolex version : 2.7.4

`reset()` resets the high-resolution timer.

**What actually happens**

The high-resolution timer is not reset.

**How to reproduce**

```js
const clock = lolex.install()
clock.now === 0
clock.tick(20)
clock.hrNow === 20
clock.reset()
clock.now === 0
clock.hrNow === 20
```

SimenB:
Fixes #206


fatso83:
Out as 2.7.5

PR #210, opened 2018-09-18
fix: handle floating point in hrtime

### LINKED ISSUE #207: High-resolution timer gets tripped up by floating point arithmetic
* Lolex version : 2.7.4

Advancing the clock with fractional milliseconds should give me an integer nanosecond value when I call `process.hrtime(prev)`.

**What actually happens**

The nanosecond value is a float, not an integer.

**How to reproduce**

```js
const clock = lolex.install()
clock.tick(1022.7791)
clock.hrtime([0, 20000000]) // [ 1, 2779099.999999881 ]
```

The reproduction advanced the clock by a fraction of a nanosecond. I suppose it'd be fine to round the computed nanosecond value, but I'm not sure if it should be rounded up or down. Alternatively something like https://www.npmjs.com/package/decimal.js may help to avoid the floating point arithmetic. Perhaps `tick()` should round down to the closest nanosecond, but I'm not sure if that'll help or not.

SimenB:
Fixes #207

/cc @novemberborn

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -101,7 +101,7 @@ function withGlobal(_global) {
      * % operator that also works for negative numbers
      */
     function fixedModulo(n, m) {
-        return ((n % m) + m) % m;
+        return Math.round(((n % m) + m) % m);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


fatso83:
:+1: I think this is fine. If someone ever gets a problem with round-off errors ... we'll tackle it then.

fatso83:
Out as 2.7.5

benjamingr:
Thanks for all those changes @SimenB  :)

PR #214, opened 2018-10-06
Introduce nanosecond precision ticks

### LINKED ISSUE #207: High-resolution timer gets tripped up by floating point arithmetic

fatso83:
#### Purpose 
This is a fix for #207, which dealt in round-off errors. 

Other notable changes are:
- removes `clock.hrNow`. Should not be an issue as it's an internal detail not documented in our API
- throws on negative ticks. This is a breaking API change, but I would be very surprised if it affected anyone, as this essentially means backwards time travel.
- change some test structure, replacing `if()` blocks with conditional `this.skip()`

#### Background
This PR fixes #207 by doing away with floats for representing time; opting to store milliseconds and millisecond decimals separately. We could have avoided the complications of storing two parts by opting to store all ticks as nanoseconds, but that would pose other round-off issues when adding/subtracting high-value milliseconds. Javascript is unable to represent the epoch as nanos due to too few bits for the `Number` type. See discussion in #207 for more background. The test cases found through that issue is included in the test suite.

I tried a few approaches, but this approach was the simplest one to follow and allowed me to do away with a bit of code. For a while I had problems with the behaviour that setting system time should not affect the hrtime, and I tried recreating the approach of using a separate clock for `hrtime` (which was the purpose `hrNow` served), but keeping the two clocks in sync, including the nanosecond offsets, with all the existing `updateHrTime()` callsites was just too much for my simple mind to fathom :boom: Instead I found a simpler solution which was just to record the delta between the new and old time when a new system clock is set and use that to compensate in the `hrtime()` function (see ae83b27).

closes #207

#### Verifying
1. clone the repo and checkout the branch
1. `npm install && npm test`
1. `npm run test-cloud` (if you have sauce labs credentials)
1. `npm link` - will install this as your local `lolex`
1. Require `lolex` in your repo and run against your own code/tests

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -105,6 +105,17 @@ function withGlobal(_global) {
     }
 
     /**
+     * @returns the remaining number of nanoseconds.
+     *
+     * Example: nanoRemainer(123.456789) -> 456789
+     */
+    function nanoRemainder(msFloat){
+        var modulo = 1e6;
+        var remainder = (msFloat * 1e6) % modulo;
+        return remainder < 0 ? remainder + modulo: remainder;
+    }
+
+    /**
      * Used to grok the `now` parameter to createClock.
      * @param epoch {Date|number} the system time
      */
@@ -505,12 +516,13 @@ function withGlobal(_global) {
     };
 
     /**
-     * @param start {Date|number} the system time
+     * @param start {Date|number} the system time - non-integer values are floored
      * @param loopLimit {number}  maximum number of timers that will be run when calling runAll()
      */
     function createClock(start, loopLimit) {
-        start = start || 0;
+        start = Math.floor(getEpoch(start));
         loopLimit = loopLimit || 1000;
+        var nanos = 0;
 
         if (NativeDate === undefined) {
             throw new Error("The global scope doesn't have a `Date` object"
@@ -518,11 +530,10 @@ function withGlobal(_global) {
         }
 
         var clock = {
-            now: getEpoch(start),
-            hrNow: 0,
+            now: start,
             timeouts: {},
             Date: createDate(),
-            loopLimit: loopLimit
+            loopLimit: loopLimit,
         };
 
         clock.Date.clock = clock;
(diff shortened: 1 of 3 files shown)


fatso83:
> this still can be potentially not as precise

Good point! We discussed this initially and agreed on truncating anything less than a nanosecond, so I'll add a final commit. Should suffice with a `Math.floor()` in [`nanoRemainder`](https://github.com/sinonjs/lolex/pull/214/files#diff-3931e3e3208b32fabec68afaa455f300R99).

fatso83:
Truncation of sub-nanosecond values in place.

test/lolex-test.js
```diff
@@ -114,6 +114,93 @@ describe("issue sinon#1852", function () {
...
+        assert.equals(clock.now, -2);
+    });
+
+    it("should floor start times", function () {
+        clock = lolex.install({now: 1.2});
+        assert.equals(clock.now, 1);
+    });
+
+    it("should floor negative start times", function () {
+        clock = lolex.install({now: -1.2});
```
SimenB: does it make sense with negative start times?

src/lolex-src.js
```diff
@@ -531,6 +534,30 @@ function withGlobal(_global) {
...
+                if ( prev[1] > 1e9 ) {
+                    throw new TypeError("Number of nanoseconds can't exceed a billion");
+                }
+
+                var oldSecs = prev[0];
+                var nanoDiff = remainderInNanos - prev[1];
+                var secDiff = secsSinceStart - oldSecs;
+
+                if (nanoDiff < 0) {
+                    nanoDiff += 1 * 1e9;
```
SimenB: why the `1`?

test/lolex-test.js
mroderick: Yes, that just means time before 1970 :)

test/lolex-test.js
```diff
@@ -114,6 +114,50 @@ describe("issue sinon#1852", function () {
     });
 });
 
+describe.only("issue #207 - nanosecond round-off errors on high-res timer", function () {
```
mroderick: That's an exclusive test. If we land on this commit using `git bisect`, we might mistakenly think all is well. Will you remove this from history, or are you planning to use "Squash and merge" when merging this PR?

src/lolex-src.js
```diff
@@ -626,6 +626,10 @@ function withGlobal(_global) {
             var nanosTotal = nanos + remainder;
             var tickTo = clock.now + ms;
 
+            if (msFloat < 0) {
+                throw new TypeError("Only positive ticks are supported");
```
mroderick: The error message is a bit misleading. Are zero ticks allowed?
If not, shouldn't the conditional be `msFloat <= 0`?

benjamingr: I'm also fine with ``.tick` ms value must be called with a value >= 0` or something like that.

src/lolex-src.js
fatso83: Do you mean "Why do I add a billion" or why do I use `1 * 1e9` instead of just `1e9`. In case of the latter, no particular reason. I can change it.

test/lolex-test.js
fatso83: Didn't see that. I thought this would make the most sense as a merge commit. Can try to fix it.

src/lolex-src.js
fatso83: Changing to `Negative ticks are not supported`

mroderick:
I have no further comments, would like to see at least one other 👍 before we merge this though

src/lolex-src.js
SimenB: The latter 🙂

fatso83:
@SimenB  As in, if we can come up with other breaking changes we would like to see implemented before releasing a new major?

Or, can I think if any other breaking changes in this PR?

fatso83:
Thanks for checking, @novemberborn! Good to go, I assume.

fatso83:
Released Lolex 3

Issue #213, opened 2018-10-04
[feat] Way to check that there are zero waiting timers


thejohnfreeman:
It'd be nice to have a method that can tell me if there are any waiting timers, so I can assert in tests that every timer has run or been cleaned up.

Reading the code, I see that clocks have a `timers` property which is an object mapping timer IDs to timers. I could check that that object has zero own properties, but it is not a documented part of the public API, so I'm at risk of breaking changes.

Another option would have `runAll` return the number of timers it ran, so that I could assert at the end of my tests that number is zero, but it is returning `clock.now` in the current version. That does not appear to be a documented part of the API, but changing it risks breaking users mistakenly depending on it.

What do you think?

SimenB:
Dupe of #163.

We need this Jest as well, FWIW.

fatso83:
I think we would love a PR for such a feature :wink: Check out #163 for some background on what we need.

thejohnfreeman:
I'll work on it this weekend. Just some docs and test(s) on the return value of `.uninstall`?

SimenB:
We shouldn't have to uninstall to get the count. I might want to check e.g. remaining timers per run of my function or something

benjamingr:
> We shouldn't have to uninstall to get the count. I might want to check e.g. remaining timers per run of my function or something

Did anyone actually ever ask for that for a real use case? Not saying that it's not important - only that I haven't seen a request.

SimenB:
Facebook has a test for debouncing or something like that, which checks how many timers are scheduled after calling a function. Not sure if the uninstall would be enough in that case, actually

PR #215, opened 2018-10-07
Add countTimers method

### LINKED ISSUE #163: Count of active timers
Hello,

I would like to find out how many timers is actively running, but I can't find any mention of it in the API reference.

My use case is that I'm writing a test for a destructor function and I would like to make sure that all timers got cancelled after the function execution so there would be no memory leaks.

Currently I'm using this hotfix `Object.keys(clock.timers || {}).length`, but it seems quite verbose and dependent on the inner implementation.

thejohnfreeman:
Fixes #163 and #213 by adding a method to count the waiting timers. I (and I believe others: Facebook, @radarfox, @SimenB, ?) want this method to check for "timer leaks" in our tests.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -200,6 +200,11 @@ available in both browser & node environments.
 
 Cancels the callback scheduled by the provided id.
 
+### `clock.countTimers()`
+
+Returns the number of waiting timers. This can be used to assert that a test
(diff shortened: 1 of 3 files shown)


fatso83:
Thanks for the work. Released as part of lolex 3.0

PR #217, opened 2018-10-16
add support for queueMicrotask

benjamingr:

Adds support for queueMicrotask added in Node.js 11 https://github.com/nodejs/node/pull/22951

Browsers should also eventually add this. 

This is also not urgent since Node.js 11 isn't even out yet.

#### Solution  - optional

This is essentially a nextTick that drops arguments which we already support.


SimenB:
Any plans to release this? Node 11 is out 🙂

benjamingr:
@SimenB just need to do a release really. @fatso83 ?

PR #228, opened 2018-11-17
Upgrade cached-path-relative

mroderick:
See https://nvd.nist.gov/vuln/detail/CVE-2018-16472

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -880,9 +880,9 @@
       }
     },
     "cached-path-relative": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/cached-path-relative/-/cached-path-relative-1.0.1.tgz",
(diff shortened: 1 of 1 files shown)


SimenB:
any reason they are not semver range? seeing as we have a lock file, there's no reason to lock down our dependencies in package.json

mroderick:
> any reason they are not semver range? seeing as we have a lock file, there's no reason to lock down our dependencies in package.json

No reason that I can recall. Since we have a lock file (assuming that developers install using `npm ci`), it won't make any difference.

However we should probably change it to use ranges, so users will get up to date dependencies when they install. Would you like to make a PR for this?

mroderick:
This PR has been updated to only update cached-path-relative, as a full upgrade of browserify causes builds to fail

Issue #232, opened 2019-02-12
requiring lolex prints a warning


ehmicky:
`require('lolex')` prints the following warning:

```
(node:30019) ExperimentalWarning: queueMicrotask() is experimental.
```

* Lolex version : `3.1.0`
* Environment   : Node `11.9.0`, Ubuntu `18.10`

**What did you expect to happen?**
There should be no warnings printed on the console when `queueMicrotask()` is not used by the calling code. This could for example be done by wrapping `timers.queueMicrotask()` so that `global.queueMicrotask` is only accessed when `timers.queueMicrotask()` is called.

Experimental warnings can be turned off with `--no-warnings` but this is inconvenient as it removes other potentially useful warnings. 

Why suppressing this warning matters: some users (like me) might have precise expectations on what's printed on the console, if console output is being parsed, for example for logging purpose. Also some users might see this warning and without additional context think it was triggered from their code, making them debug a non-issue.

**What actually happens**
There is a warning printed on console.

**How to reproduce**
See above.

benjamingr:
@shdq

shdq:
I'll take a look at this.

benjamingr:
@shdq thanks, a quick fix would be not to shim `enqueueMicrotask` by default.

ehmicky:
Also for the detection: instead of `typeof global.queueMicrotask === 'function'` (which prints a warning), something like `global.hasOwnProperty('queueMicrotask')` does not print a warning.

SimenB:
In general, should updating the default shimmed functions be considered a breaking change? So don't shim new stuff by defsult, then for new majors activate everything?

benjamingr:
Good idea @SimenB

SneakyMax:
Some console things, for example ava's default reporter, start to break a little if there are any console logs or warnings during operation, and because each ava node process ends up logging a new warning, the output gets rather messy because of this issue. I've turned off node warnings with `NODE_NO_WARNINGS=1`, but obviously I'd like to turn those back on.

ehmicky:
@SneakyMax A pending PR #233 exists. Until this PR is merged, if you want to only hide the warning printed by Lolex, you can use [`log-process-errors`](https://github.com/ehmicky/log-process-errors) to remove this warning during Ava tests:

```js
// Those two lines must be run before `require('lolex')` is first called
const logProcessErrors = require('log-process-errors')
logProcessErrors({
  level: { 
    warning({ value }) {
[...]

PR #233, opened 2019-02-16
[WIP] queueMicrotask() ExperimentalWarning in Node

shdq:
Hello, guys.

I did a quick fix for #232 not faking this method by default like `nextTick`, but have a couple of questions.

`--trace-warnings` point on `_global.queueMicrotask` usage in the code.

There are two places in the code:

1. Feature detection (I fixed in @ehmicky way). Not sure about that, because it gives `true` when property value is `null` or `undefined`. What do you think?
2. And when it stores in `lolex.timers`. What is the purpose of storing all the timers methods that supported by particular environment there instead of storing only just list of them? As I got from the code `hijackMethod` assign `_queueMicrotask` to `_global.queueMicrotask`, but not to `timers.queueMicrotask`. I might be incorrect.

Now it is giving warnings only with `lolex.install({ toFake: ["queueMicrotask"] })`.

I'm going to add some tests for this.

@benjamingr I fixed queueMicrotask() test "runs with timers and before them", which was failed, I think it was the typo.

Any comments?

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -33,7 +33,7 @@ function withGlobal(_global) {
     var nextTickPresent = (_global.process && typeof _global.process.nextTick === "function");
     var performancePresent = (_global.performance && typeof _global.performance.now === "function");
     var hasPerformancePrototype = (_global.Performance && (typeof _global.Performance).match(/^(function|object)$/));
-    var queueMicrotaskPresent = (typeof _global.queueMicrotask === "function");
+    var queueMicrotaskPresent = (_global.hasOwnProperty("queueMicrotask"));
     var requestAnimationFramePresent = (
         _global.requestAnimationFrame && typeof _global.requestAnimationFrame === "function"
     );
@@ -493,8 +493,9 @@ function withGlobal(_global) {
     if (requestAnimationFramePresent) {
         timers.requestAnimationFrame = _global.requestAnimationFrame;
     }
+
     if (queueMicrotaskPresent) {
-        timers.queueMicrotask = _global.queueMicrotask;
+        timers.queueMicrotask = true;
     }
 
     if (cancelAnimationFramePresent) {
@@ -897,7 +898,9 @@ function withGlobal(_global) {
 
         if (clock.methods.length === 0) {
             // do not fake nextTick by default - GitHub#126
-            clock.methods = keys(timers).filter(function (key) {return key !== "nextTick";});
+            clock.methods = keys(timers).filter(function (key) {
+                return key !== "nextTick" && key !== "queueMicrotask";
+            });
         }
 
         for (i = 0, l = clock.methods.length; i < l; i++) {
--- a/test/lolex-test.js
+++ b/test/lolex-test.js
@@ -2545,7 +2545,7 @@ describe("lolex", function () {
         });
         it("runs with timers and before them", function () {
             var last = "";
-            clock.runMicrotasks(function () {
+            clock.queueMicrotask(function () {
                 called = true;
                 last = "tick";
             });


src/lolex-src.js
```diff
@@ -493,8 +493,9 @@ function withGlobal(_global) {
     if (requestAnimationFramePresent) {
         timers.requestAnimationFrame = _global.requestAnimationFrame;
     }
+
     if (queueMicrotaskPresent) {
-        timers.queueMicrotask = _global.queueMicrotask;
+        timers.queueMicrotask = true;
```
benjamingr: Why is this set to true?

benjamingr:
> 1. What do you think? 

I think the property check fix is fine since there is no property called enqueueMicrotask on the process object otherwise

ehmicky:
@shdq thanks for taking some time fixing my issue!

About `hasOwnProperty()` you are right it does not work when the property is explicitely set to `undefined`. I am not sure whether some libraries or environments might do this. Would chaining `global.hasOwnProperty("queueMicrotask") && typeof global.queueMicrotask === "function"` fix this?

src/lolex-src.js
shdq: It gives another warning for just `var lolex = require("lolex");`.
Assigned native method not used later. So it's just for "yes, this feature supported". And actually we can remove it at all, because we filtering this method later inside install function to prevent faking by default. Correct me, please, if I'm wrong.

benjamingr: I think this is here for restoring the original queueMicrotask later

shdq: I thought so, but during `uninstall` function it restores original method from `_queueMicrotask` which was assigned inside `hijackMethod` from `target[method]` (`_global[queueMicrotask` in our case). `lolex.timers` not involved in this at all.

Example code on node 11.9.0:
```
const lolex = require("lolex");
var clock = lolex.install({ toFake: ["setTimeout", "queueMicrotask"] });

function microtask() {
  console.log("microtask");
}
function task() {
  console.log("task");
}

setTimeout(task, 0);
queueMicrotask(microtask);

clock.tick();

clock.uninstall();

setTimeout(task, 0);
queueMicrotask(microtask);
```
Outputs
```
microtask // faked method
task // faked method
(node:27136) ExperimentalWarning: queueMicrotask() is experimental.
    at get (internal/bootstrap/node.js:771:15)
    at hijackMethod (/Users/shdq/Projects/lolex-warning/node_modules/lolex/src/lolex-src.js:486:33)
    at Object.install (/Users/shdq/Projects/lolex-warning/node_modules/lolex/src/lolex-src.js:995:9)
    at Object.<anonymous> (/Users/shdq/Projects/lolex-warning/index.js:3:19)
    at Module._compile (internal/modules/cjs/loader.js:734:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:745:10)
    at Module.load (internal/modules/cjs/loader.js:626:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:566:12)
    at Function.Module._load (internal/modules/cjs/loader.js:558:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:797:12)
microtask // native method
task // native method
```
It gives this warning on line 2, because there in hijackMethod `_queueMicrotask = _global[queueMicrotask]`;

I'm not insisting on `true`, we can even remove it at all, because this method is not going to be faked by default after this PR, but maybe it will be against whole concept of storing all the supported timers in `lolex.timers`. What do you think?

shdq:
> Either way: would chaining `global.hasOwnProperty("queueMicrotask") && typeof global.queueMicrotask === "function"` fix this? The right side would only get evaluated if the left side is `true`.

Thank you for the suggestion, when the right side will be evaluated it gives a warning in a console.

ehmicky:
@shdq Yes you're right my suggestion did not make much sense :)

boneskull:
Would love to see this merged; anything I can do to help?

benjamingr:
@boneskull sorry, had a lot on my plate lately. @fatso83 can you do a release? 

(If not I'll get one next week + hopefully the setTimeout promises thing)

ehmicky:
I tried it and it works. Thanks!

fatso83:
@benjamingr I tried doing it in a hurry on a mac I was able to loan, but `npm version patch` failed me:
```
scripts/preversion.sh 
scripts/preversion.sh: line 8: conditional binary operator expected
```

So it will have to wait until next week for me. Perhaps @mroderick ?

fatso83:
OK, I am having a stab at this now.

fatso83:
Out as `lolex@4"

PR #240, opened 2019-05-02
Return empty arrays for performance.getEntries, other relevant methods

flotwig:
#### Purpose (TL;DR)
Fix issues caused by applications which expect `performance.getEntries`, `performance.getEntriesByName`, and `performance.getEntriesByType` to return arrays (such as Facebook SDK - downstream issue: cypress-io/cypress#3625)

Right now, lolex clobbers these with `() => undefined`. Just using `() => []` instead would fix this.

MDN page for Performance functions: https://developer.mozilla.org/en-US/docs/Web/API/Performance

Only 5 have non-undefined return types. The other 2 (besides `getEntries.*`) are:
* `now()` - already done by lolex
* `toJSON()` - not handled, not sure if there's a good way to mock this value

#### Solution 

Just have those 3 methods return `[]` when `lolex.install()` is called, instead of `undefined`

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -850,7 +850,7 @@ function withGlobal(_global) {
                 Object
                     .getOwnPropertyNames(proto)
                     .forEach(function (name) {
-                        if (name.slice(0, 10) === "getEntries") {
+                        if (name.indexOf("getEntries") === 0) {
                             // match expected return type for getEntries functions
                             clock.performance[name] = NOOP_ARRAY;
                         } else {
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -27,6 +27,7 @@ function withGlobal(_global) {
     // see https://github.com/cjohansen/Sinon.JS/pull/436
 
     var NOOP = function () { return undefined; };
+    var NOOP_ARRAY = function () { return []; };
     var timeoutResult = _global.setTimeout(NOOP, 0);
     var addTimerReturnsObject = typeof timeoutResult === "object";
     var hrtimePresent = (_global.process && typeof _global.process.hrtime === "function");
@@ -849,7 +850,12 @@ function withGlobal(_global) {
                 Object
                     .getOwnPropertyNames(proto)
                     .forEach(function (name) {
-                        clock.performance[name] = NOOP;
+                        if (name.slice(0, 10) === "getEntries") {
+                            // match expected return type for getEntries functions
+                            clock.performance[name] = NOOP_ARRAY;
+                        } else {
+                            clock.performance[name] = NOOP;
+                        }
                     });
             }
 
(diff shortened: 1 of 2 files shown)


test/lolex-test.js
```diff
@@ -2084,6 +2084,24 @@ describe("lolex", function () {
...
+
+                this.clock = lolex.install();
+
+                assert.equals(performance.getEntries(), []);
+                assert.equals(performance.getEntriesByName(), []);
+                assert.equals(performance.getEntriesByType(), []);
+
+                this.clock.uninstall();
+
+                delete Performance.prototype.getEntries;
```
fatso83: I am just wondering about the cleanup bit. Will the original functions now be "restored"?

flotwig: Just added a test that the original functions are restored, it does appear to work.

I also have a test for Cypress that runs in actual Chrome that shows the restoration working IRL: https://github.com/cypress-io/cypress/pull/4108/files#diff-c0a131a0b58412db68e166e9d7a16383R75

src/lolex-src.js
```diff
@@ -849,7 +850,12 @@ function withGlobal(_global) {
                 Object
                     .getOwnPropertyNames(proto)
                     .forEach(function (name) {
-                        clock.performance[name] = NOOP;
+                        if (name.slice(0, 10) === "getEntries") {
```
SimenB:
```suggestion
                        if (name.indexOf("getEntries") === 0) {
```

test/lolex-test.js
SimenB: I _think_ he asked about the restoration within tests since you override and `delete` from `prototype`

flotwig: Hmm, I don't see what you mean. Is there something else that needs to be tested?

SimenB: dunno, maybe something like

```js
const origGetEntries = Performance.prototype.getEntries;
const origGetEntriesByName = Performance.prototype.getEntriesByName;
const origGetEntriesByType = Performance.prototype.getEntriesByType; =
```

then restore them instead of the `delete`? not sure. If the `Performance` object is already a fake, then it doesn't matter 🙂

flotwig: I could be wrong, but it looks like `lolex` just uses the prototype to know what methods to override on the global `performance`, so I don't think anything additional needs to be tested here.

SimenB: @fatso83 is this ready?

fatso83: Sure. If you and Ben thinks this is fine, go ahead

benjamingr: I think this is fine :]

benjamingr:
I can do a release next week when I'm back from DevDays

flotwig:
@benjamingr Ping, would be awesome to see this released so we can incorporate this fix downstream :)

fatso83:
doing it

PR #241, opened 2019-05-11
fix: count microtasks in countTimers

### LINKED ISSUE #238: Include ticks in countTimers

* Lolex version : 4.0.1
* Environment   : Node 10
* Example URL   : N/A
* Other libraries you are using: N/A

See https://github.com/facebook/jest/pull/8139 for justification. Tl;dr; is that a timer count should include all scheduled timers _and friends_, not just timers.

**What actually happens**

`clock.nextTick` does not contribute to the count

**How to reproduce**

```js
const lolex = require("lolex");

const clock = lolex.createClock();

clock.setTimeout(() => {}, 5);
clock.nextTick(() => {});

console.log(clock.countTimers()); // 1
```

SimenB:

Fixes #238

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -645,7 +645,7 @@ function withGlobal(_global) {
         };
 
         clock.countTimers = function countTimers() {
-            return Object.keys(clock.timers || {}).length;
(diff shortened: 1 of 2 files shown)


SimenB:
Release as well? 🙏

PR #244, opened 2019-05-27
Support hrtime.bigint()

### LINKED ISSUE #235: Support hrtime.bigint()

* Lolex version : 3.1.0
* Environment   : Node.js 10.15.3
* Example URL   : N/A

My code uses [`process.hrtime.bigint()`](https://nodejs.org/api/process.html#process_process_hrtime_bigint). When stubbing `process.hrtime` using Lolex, this method is no longer available.

On platforms that don't provide this function, I'd expect it not be made available by Lolex either.

**What actually happens**

It returns the stubbed time, as a BigInt.

**How to reproduce**
> Describe *with code* how to reproduce the faulty behaviour
> or link to code on JSBin or similar

```console
$ node                                                                                                                                            > process.hrtime.bigint()
39069092115824n
> require('lolex').install({toFake: ['hrtime']})
{ now: 0, … }
> process.hrtime.bigint()
TypeError: process.hrtime.bigint is not a function
```

novemberborn:
Fixes #235.

I didn't think it was relevant to mention this in the documentation.

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -31,6 +31,7 @@ function withGlobal(_global) {
     var timeoutResult = _global.setTimeout(NOOP, 0);
     var addTimerReturnsObject = typeof timeoutResult === "object";
     var hrtimePresent = (_global.process && typeof _global.process.hrtime === "function");
+    var hrtimeBigintPresent = (hrtimePresent && typeof _global.process.hrtime.bigint === "function");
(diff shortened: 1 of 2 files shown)


benjamingr:
Nice work, this looks good.

Need to do a release :]

novemberborn:
Thanks!

PR #245, opened 2019-06-20
Fix support for replacing the JSDOM performance field

### LINKED ISSUE #243: TypeError: Cannot set property performance of #<Object> which has only a getter
* Lolex version : 4.0.1
* Environment   : node@10.15.1
  - jsdom@15.1.0

This is an issue that was raised before as #197. That issue links to code changes that would probably fix this issue.  However, those changes never appeared to have made it to `master`.

Lolex to initialize without error

**What actually happens**
I get the stack trace:
```
/Users/cdow2/workspace/lolex-test/node_modules/lolex/src/lolex-src.js:451
                           ^

TypeError: Cannot set property performance of #<Object> which has only a getter
    at hijackMethod (/Users/cdow2/workspace/lolex-test/node_modules/lolex/src/lolex-src.js:451:28)
    at Object.install (/Users/cdow2/workspace/lolex-test/node_modules/lolex/src/lolex-src.js:923:17)
    at Object.<anonymous> (/Users/cdow2/workspace/lolex-test/index.js:35:21)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
    at startup (internal/bootstrap/node.js:283:19)
```

**How to reproduce**
I'm trying to use the jsdom setup for enzyme described [here](https://github.com/airbnb/enzyme/blob/master/docs/guides/jsdom.md).  This can be reproduced by creating an node project with the following files:

<details>
    <summary>package.json</summary>

```json
{
  "name": "lolex-test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "start": "node index.js"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "jsdom": "^15.1.0",
    "lolex": "^4.0.1"
  }
}
```
</details>

<details>
    <summary>index.js</summary>

```js

//
// Setup described in: https://github.com/airbnb/enzyme/blob/master/docs/guides/jsdom.md
//
const { JSDOM } = require('jsdom');

const jsdom = new JSDOM('<!doctype html><html><body></body></html>');
const { window } = jsdom;

function copyProps(src, target) {
  Object.defineProperties(target, {
    ...Object.getOwnPropertyDescriptors(src),
    ...Object.getOwnPropertyDescriptors(target),
  });
}

global.window = window;
global.document = window.document;
global.navigator = {
  userAgent: 'node.js',
};
global.requestAnimationFrame = function (callback) {
  return setTimeout(callback, 0);
};
global.cancelAnimationFrame = function (id) {
  clearTimeout(id);
};
copyProps(window, global);

//
// Main code
//
const lolex = require('lolex');
clock.uninstall();
```
</details>

Then run the following commands:
```
npm install
npm start
```

cdow:

Fix #243 by changing the way the `performance` field gets overwritten

JSDOM has a read only `performance` field.  So, when JSDOM is configured to provide [global values](https://github.com/airbnb/enzyme/blob/master/docs/guides/jsdom.md), lolex will error on init. This is because it can't overwrite the global `performance` field.

### DIFF
--- a/integration-test/lolex-integration-test.js
+++ b/integration-test/lolex-integration-test.js
@@ -57,3 +57,83 @@ describe("withGlobal", function () {
         clock.uninstall();
     });
 });
+
+describe("globally configured browser objects", function () {
+    var withGlobal, originalDescriptors;
+
+    // Used to support Node 6 which doesn't support Object.getOwnPropertyDescriptors
+    function getOwnPropertyDescriptors(obj) {
+        var propertyNames = Object.getOwnPropertyNames(obj);
+        var result = {};
+        propertyNames.forEach(function (name) {
+            var descriptor = Object.getOwnPropertyDescriptor(obj, name);
+            result[name] = descriptor;
+        });
+
+        return result;
+    }
+
+    // We use a set up function instead of beforeEach to avoid Mocha's check leaks detector
+    function setUpGlobal() {
+        // Configuration taken from from here https://github.com/airbnb/enzyme/blob/master/docs/guides/jsdom.md
+        var dom = new jsdom.JSDOM("<!doctype html><html><body></body></html>");
+        var window = dom.window;
+
+        function copyProps(src, target) {
+            originalDescriptors = getOwnPropertyDescriptors(target);
+            Object.defineProperties(target, getOwnPropertyDescriptors(src));
+            Object.defineProperties(target, originalDescriptors);
+        }
+
+        global.window = window;
+        global.document = window.document;
+        global.navigator = {
+            userAgent: "node.js"
+        };
+        global.requestAnimationFrame = function (callback) {
+            return setTimeout(callback, 0);
+        };
+        global.cancelAnimationFrame = function (id) {
+            clearTimeout(id);
+        };
+        copyProps(window, global);
+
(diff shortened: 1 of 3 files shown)


fatso83:
This is great work. My only issue with it is that I have no way of seeing if it _does the right thing_.  Could you please add a test to `integration-test/lolex-integration-test.js` that would break before your fix? Perhaps even rebase the commits once this has been done and change the ordering so that it's easily verifiable by just checkout out the second-to-last commit?

cdow:
Test added

fatso83:
Thank yoy!

fatso83:
This broke massively on other browsers. This was my fault, as I did not test this in other browsers (which is only done on the `master` branch). I had to revert the fix.

fatso83:
https://travis-ci.org/sinonjs/lolex/jobs/549635521

Issue #246, opened 2019-07-19
`Date()` should return a string, not an object, in Lolex v4.1.0


moll:
Hey,

I believe Lolex deviates from the JavaScript standard behavior with regards to what `Date()` returns. I noticed this through a bug in my code that passed all tests, yet failed in production: I was creating a new date without `new` — `Date(2015, 5, 18)`. In tests, that was equivalent to `new Date(..)`, but without mocking, it returns the current time as a string — an entirely different result. It seems to be an issue already since the times of Sinon v1.17.7 and possibly earlier. Here's an example from RunKit https://runkit.com/embed/delfomea61tm with Lolex v4.1.

```js
require("lolex").install()
Date(2015, 5, 18) // => Thu Jun 18 2015 00:00:00 GMT+0000 (Coordinated Universal Time)
```

fatso83:
Verified! The standard is strange :)

moll:
I think the stale bot is too eager here.

PR #247, opened 2019-07-21
Fix bug with returning date as string when calling Date as function.

life777:
Fix issue #246 by returning string date when Date constructor is called as function

Based on the ECMAScript standard if Date constructor is called as function (without 'new' keyword) it should return a date string. [See](https://www.ecma-international.org/ecma-262/#sec-date-year-month-date-hours-minutes-seconds-ms)

#### Solution

We should check whether ClockDate function is called with 'new' keyword or without and return Date object or convert it to string. Also a few tests should be fixed.

### DIFF
--- a/lolex.js
+++ b/lolex.js
@@ -172,12 +172,10 @@ function withGlobal(_global) {
     }
 
     function createDate() {
-        function ClockDate(year, month, date, hour, minute, second, ms) {
+        function createNativeDate(year, month, date, hour, minute, second, ms) {
             // Defensive and verbose to avoid potential harm in passing
             // explicit undefined when user does not pass argument
             switch (arguments.length) {
-                case 0:
-                    return new NativeDate(ClockDate.clock.now);
                 case 1:
                     return new NativeDate(year);
                 case 2:
@@ -193,6 +191,24 @@ function withGlobal(_global) {
                 default:
                     return new NativeDate(year, month, date, hour, minute, second, ms);
             }
+        }
+
+        function ClockDate() {
+            var args = Array.prototype.slice.call(arguments);
+
+            if (args.length === 0) {
+                args.push(ClockDate.clock.now);
+            }
+
+            var date = createNativeDate.apply(this, args);
+
+            // if Date is called as a constructor with 'new' keyword
+            if (this instanceof ClockDate) {
+                return date;
+            }
+
+            // if Date called as function - ecma-262/#sec-date-year-month-date-hours-minutes-seconds-ms
+            return date.toString();
         }
 
         return mirrorDateProperties(ClockDate, NativeDate);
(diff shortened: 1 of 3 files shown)


lolex.js
```diff
@@ -172,12 +172,10 @@ function withGlobal(_global) {
     }
 
     function createDate() {
-        function ClockDate(year, month, date, hour, minute, second, ms) {
+        function createNativeDate(year, month, date, hour, minute, second, ms) {
```
fatso83: Revert all changes to `lolex.js`. This is changed only by build scripts.

src/lolex-src.js
```diff
@@ -193,6 +191,24 @@ function withGlobal(_global) {
             }
         }
 
+        function ClockDate() {
+            var args = Array.prototype.slice.call(arguments);
+
+            if (args.length === 0) {
```
fatso83: I don't understand why this case needed to be moved out? Is there something particular about this case that would cause different behavior when left in the switch case?

src/lolex-src.js
```diff
@@ -170,12 +170,10 @@ function withGlobal(_global) {
     }
 
     function createDate() {
-        function ClockDate(year, month, date, hour, minute, second, ms) {
+        function createNativeDate(year, month, date, hour, minute, second, ms) {
             // Defensive and verbose to avoid potential harm in passing
             // explicit undefined when user does not pass argument
             switch (arguments.length) {
-                case 0:
```
fatso83: Keep this case. This is pretty easy to read and keeps all the logic that deals with the number of arguments in the same place, instead of spreading the logic around.

src/lolex-src.js
```diff
@@ -193,6 +191,24 @@ function withGlobal(_global) {
...
+            }
+
+            var date = createNativeDate.apply(this, args);
+
+            // if Date is called as a constructor with 'new' keyword
+            if (this instanceof ClockDate) {
+                return date;
+            }
+
+            // if Date called as function - ecma-262/#sec-date-year-month-date-hours-minutes-seconds-ms
```
fatso83: Since we are actually explicitly targetting version 5.1, it's worth adding that this has remained unchanged from version 5.1, where it is mentioned in [section 15.9.2](https://www.ecma-international.org/publications/files/ECMA-ST-ARCH/ECMA-262%205.1%20edition%20June%202011.pdf).
```suggestion
            // the Date constructor called as a function, ref Ecma-262 Edition 5.1, section 15.9.2. 
            // This remains so in the 10th edition of 2019 as well.
```

test/lolex-test.js
```diff
@@ -1622,10 +1622,10 @@ describe("lolex", function () {
             assert(Date.prototype.isPrototypeOf(date));
         });
```
fatso83: Please don't change the existing tests to test for the string. Revert the changes to the test file and change them to create a new instance by adding `new `. As in `new this.clock.Date()`

Then add a new test for the behaviour you are testing for. Which is to test that when _not_ called with new, it returns a string. You don't need to test for all the cases; I'd say it suffices to test for a single case to know the logic is sound.

life777: Hi, thanks. Actually tests for each set of parameters for calling Date as a function were created before, I just fixed result. You can take a look there are two tests for each set of parameters in master:

Screenshot from 2019-08-22 16-14-52

Should I remove those tests?

src/lolex-src.js
life777: Thanks. I couldn't remember. You are right it shouldn't be a separate case.

lolex.js
life777: Done

src/lolex-src.js
life777: Done

src/lolex-src.js
life777: Done

test/lolex-test.js
SimenB: Yeah, I think the old tests were just _wrong_, no?

test/lolex-test.js
```diff
@@ -1687,95 +1687,58 @@ describe("lolex", function () {
             assert.equals(fakeDate.getTime(), date.getTime());
         });
 
-        it("returns regular date when calling with timestamp", function () {
-            var date = new Date();
-            var fakeDate = this.clock.Date(date.getTime());
```
SimenB: in the cases of these deleted tests, maybe just change `this.clock.Date(...)` to `new this.clock.Date(...)`? Unless the exact same tests are somewhere else as well

fatso83:
I had to manually fix up and clean it up a bit (like reverting changes to the distribution file), so this ended up being squashed into a single commit: 1adc76e. Will be being pushed to NPM shortly :)

Thanks for the work!

PR #87, opened 2016-11-09
Fix sinonjs/sinon#1143: Only overwrite globals when running in IE

mroderick:
This is a PR for **master**, it is a solution for https://github.com/sinonjs/sinon/issues/1143

The PR fixes compatibility with React Native and quite possibly other libraries.

#### To verify bug

1. Run test in https://github.com/mroderick/sinon-issue-1143
1. Observe that an error occurs

```
TypeError: Date is not a constructor

  at new MessageQueue (node_modules/react-native/Libraries/Utilities/MessageQueue.js:62:26)
  at Object.<anonymous> (node_modules/react-native/Libraries/BatchedBridge/BatchedBridge.js:17:19)
  at Object.<anonymous> (node_modules/react-native/Libraries/Utilities/RCTLog.js:14:19)
  at setUpConsole (node_modules/react-native/Libraries/JavaScriptAppEngine/Initialization/InitializeJavaScriptAppEngine.js:54:1)
```

#### To verify solution

1. Check out this PR, switch into that folder
1. `npm link`
1. Switch back to your `sinon-issue-1143` folder that you just used above
1. `cd node_modules/sinon/`
1. `npm link lolex`
1. `cd -`
1. `npm test`
1. Observe that test test passes

PR #248, opened 2019-08-03
Fix jsdom 2

### LINKED ISSUE #243: TypeError: Cannot set property performance of #<Object> which has only a getter

cdow:

Fix #243 by changing the way the `performance` field gets overwritten.  This time without breaking some browsers.

JSDOM has a read only `performance` field.  So, when JSDOM is configured to provide [global values](https://github.com/airbnb/enzyme/blob/master/docs/guides/jsdom.md), lolex will error on init. This is because it can't overwrite the global `performance` field.

### DIFF
--- a/integration-test/lolex-integration-test.js
+++ b/integration-test/lolex-integration-test.js
@@ -57,3 +57,83 @@ describe("withGlobal", function () {
         clock.uninstall();
     });
 });
+
+describe("globally configured browser objects", function () {
+    var withGlobal, originalDescriptors;
+
+    // Used to support Node 6 which doesn't support Object.getOwnPropertyDescriptors
+    function getOwnPropertyDescriptors(obj) {
+        var propertyNames = Object.getOwnPropertyNames(obj);
+        var result = {};
+        propertyNames.forEach(function (name) {
+            var descriptor = Object.getOwnPropertyDescriptor(obj, name);
+            result[name] = descriptor;
+        });
+
+        return result;
+    }
+
+    // We use a set up function instead of beforeEach to avoid Mocha's check leaks detector
+    function setUpGlobal() {
+        // Configuration taken from from here https://github.com/airbnb/enzyme/blob/master/docs/guides/jsdom.md
+        var dom = new jsdom.JSDOM("<!doctype html><html><body></body></html>");
+        var window = dom.window;
+
+        function copyProps(src, target) {
+            originalDescriptors = getOwnPropertyDescriptors(target);
+            Object.defineProperties(target, getOwnPropertyDescriptors(src));
+            Object.defineProperties(target, originalDescriptors);
+        }
+
+        global.window = window;
+        global.document = window.document;
+        global.navigator = {
+            userAgent: "node.js"
+        };
+        global.requestAnimationFrame = function (callback) {
+            return setTimeout(callback, 0);
+        };
+        global.cancelAnimationFrame = function (id) {
+            clearTimeout(id);
+        };
+        copyProps(window, global);
+
+        withGlobal = lolex.withGlobal(global);
+    }
+
+    function tearDownGlobal() {
+        var originalDescriptorNames = Object.keys(originalDescriptors);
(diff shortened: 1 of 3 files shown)


integration-test/lolex-integration-test.js
```diff
@@ -57,3 +57,83 @@ describe("withGlobal", function () {
         clock.uninstall();
     });
 });
+
+describe("globally configured browser objects", function () {
+    var withGlobal, originalDescriptors;
+
+    // Used to support Node 6 which doesn't support Object.getOwnPropertyDescriptors
```
fatso83: I'll try to fix the issues in #87 so this won't be needed.

But thank you for good comments!

integration-test/lolex-integration-test.js
```diff
@@ -57,3 +57,83 @@ describe("withGlobal", function () {
...
+        };
+        global.requestAnimationFrame = function (callback) {
+            return setTimeout(callback, 0);
+        };
+        global.cancelAnimationFrame = function (id) {
+            clearTimeout(id);
+        };
+        copyProps(window, global);
+
+        withGlobal = lolex.withGlobal(global);
```
fatso83: :+1:

PR #270, opened 2019-10-09
Use `--no-detect-globals` to bundle and test lolex

mantoni:

Our own global check `global || window` only happened to work because browserify wrapped the whole thing with the correct check. With this fix, browserify will not define `global` itself.

Note that there is a very ugly hack in the top of the test file now that can / should be removed once nise and sinon received similar updates.

See https://github.com/mantoni/mochify.js/pull/209 and linked issues.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -615,9 +615,9 @@
       "dev": true
     },
     "brout": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/brout/-/brout-1.2.0.tgz",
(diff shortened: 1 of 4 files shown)


mantoni:
Who can cut a release for this? I'm slightly unsure if this should be a patch or a minor 🤔

mantoni:
Released in `v5.0.1`.

mantoni:
@benadamstyles This is a bug, thanks for reporting! Would it be possible for you to verify that the provided patch actually works? You can copy the bundled source from jsbin here: https://jsbin.com/godesuduxo/edit?html,js,console

mantoni:
Fixed in `v5.1.1`.

PR #271, opened 2019-10-10
Upgrade eslint, add prettier

mroderick:
This PR replaces #262, as dependabot doesn't know how to upgrade when new dependencies are needed.

* Upgrade and add related plugins and configs
* Fix/disable lint violations

### DIFF
--- a/integration-test/lolex-integration-test.js
+++ b/integration-test/lolex-integration-test.js
@@ -21,23 +21,23 @@ var assert = require("@sinonjs/referee-sinon").assert;
 var lolex = require("../src/lolex-src");
 var sinon = require("@sinonjs/referee-sinon").sinon;
 
-describe("withGlobal", function () {
+describe("withGlobal", function() {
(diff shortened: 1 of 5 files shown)


fatso83:
@mroderick Merging this with all the prettier changes made a mess of #237. Could we revert this, merge #237 and then re-merge (a rebased version of) this? I tried cleaning up by quickly resetting the branch, but it's protected (yeah, yeah, yeah), so I couldn't do this without anyone noticing ...

mroderick:
> @mroderick Merging this with all the prettier changes made a conflicting mess out of #237 (which no longer had any conflicts). Could we revert this, merge #237 and then re-merge (a rebased version of) this? I tried cleaning up by quickly resetting the branch, but it's protected (yeah, yeah, yeah), so I couldn't do this without anyone noticing ...

Sure, go ahead!

fatso83:
Ha, seems @dominykas beat me to it, and actually fixed the conflicts.

PR #273, opened 2019-10-21
Fix global ReferenceError

### LINKED ISSUE #272: ReferenceError when loading lolex is browsers
Originally reported on the PR introducing the issue here: https://github.com/sinonjs/lolex/pull/270#issuecomment-544431770

* Lolex version : `5.1.0`
* Environment   : Any browser
* Example URL   : https://jsbin.com/sezegopoqa/edit?html,js,console

`lolex.createClock()` creates a clock.

**What actually happens**

A `ReferenceError` is thrown.

mantoni:

Fixes #272 which was introduced by #270.

Inspect the result of bundling this branch here: https://jsbin.com/godesuduxo/edit?html,js,console

The `global` variable is not provided by Browerify anymore. I don't fully understand why this isn't caught by test runs in browsers yet. Will need to investigate further.

#### Solution  - optional

Use the `_global` variable which is set to `window` in browsers.

### DIFF
--- a/src/lolex-src.js
+++ b/src/lolex-src.js
@@ -955,9 +955,9 @@ function withGlobal(_global) {
             return doTick(tickValue, false);
         };
 
-        if (typeof global.Promise !== "undefined") {
+        if (typeof _global.Promise !== "undefined") {
             clock.tickAsync = function tickAsync(ms) {
-                return new global.Promise(function(resolve, reject) {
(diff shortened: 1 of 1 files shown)


mantoni:
Here is the reason this wasn't caught by the tests: https://github.com/sinonjs/lolex/blob/44a8d16c94fe0f625e6a82f0d78a77fc588900bd/test/lolex-test.js#L18-L23

Issue #286, opened 2019-12-03
hrtime is affected when setSystemTime() is called multiple times


SanderDeWaal1992:
hrtime is affected when setSystemTime() is called multiple times which breaks our unit-tests. This is introduced in issue #214.

* Lolex version :5.1.1
* Environment   : 
* Example URL   : 
* Other libraries you are using:

**What did you expect to happen?**
setSystemTime() should not ever affect hrtime.
**What actually happens**
It does affect hrtime the second time it is called
**How to reproduce**
Call setSystemTime() twice

PR #287 solves this issue.

SimenB:
#287

PR #287, opened 2019-12-03
Fix setSystemTime affects hrtime if its called multiple times.

SanderDeWaal1992:
Fix issue #286


fatso83:
Thank you!

PR #292, opened 2020-01-19
support util.promisify in Node

### LINKED ISSUE #223: Promisified (with util.promisify) `setTimeout` does not work
Node.js's setTimeout promisified [does](https://github.com/nodejs/node/blob/master/lib/timers.js#L459-L463):

```js
setTimeout[internalUtil.promisify.custom] = function(after, value) {
  return new Promise((resolve) => {
    active(new Timeout(resolve, after, [value], false));
  });
};
```

This means that it does not refer to setTimeout internally - we should override `setTimeout[internalUtil.promisify.custom]` if it is available to use our fake timers and revert it on uninstall.

If anyone wants to pick this up go ahead - otherwise I will work on a fix in ~2 weeks :)

jeysal:

Fixes #223, similar to what we did for [`jest-fake-timers`](https://github.com/facebook/jest/pull/9180/files).
Noticed after starting there's already https://github.com/sinonjs/lolex/pull/227, but that seems stale and I'm not sure what some of the changes are for.
I'll do what I can to incorporate any review comments and get this through! :sweat_smile: 
If there's consensus that [storing a `global.Promise` backup](https://github.com/sinonjs/lolex/pull/227/files#r233684110) is a good idea, can do that, but would do another PR first, changing this in existing code.

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -21,6 +21,7 @@ function withGlobal(_global) {
         hrtimePresent && typeof _global.process.hrtime.bigint === "function";
     var nextTickPresent =
         _global.process && typeof _global.process.nextTick === "function";
+    var utilPromisify = _global.process && require("util").promisify;
     var performancePresent =
         _global.performance && typeof _global.performance.now === "function";
     var hasPerformancePrototype =
@@ -743,6 +744,21 @@ function withGlobal(_global) {
                 delay: timeout
             });
         };
+        if (typeof _global.Promise !== "undefined" && utilPromisify) {
+            clock.setTimeout[
+                utilPromisify.custom
+            ] = function promisifiedSetTimeout(timeout, arg) {
+                return new _global.Promise(function setTimeoutExecutor(
+                    resolve
+                ) {
+                    addTimer(clock, {
+                        func: resolve,
+                        args: [arg],
+                        delay: timeout
+                    });
+                });
+            };
+        }
 
         clock.clearTimeout = function clearTimeout(timerId) {
             return clearTimer(clock, timerId, "Timeout");
@@ -782,6 +798,22 @@ function withGlobal(_global) {
                     immediate: true
                 });
             };
+
+            if (typeof _global.Promise !== "undefined" && utilPromisify) {
+                clock.setImmediate[
+                    utilPromisify.custom
+                ] = function promisifiedSetImmediate(arg) {
+                    return new _global.Promise(function setImmediateExecutor(
+                        resolve
+                    ) {
+                        addTimer(clock, {
+                            func: resolve,
+                            args: [arg],
+                            immediate: true
+                        });
+                    });
+                };
+            }
 
             clock.clearImmediate = function clearImmediate(timerId) {
                 return clearTimer(clock, timerId, "Immediate");
--- a/test/fake-timers-test.js
+++ b/test/fake-timers-test.js
@@ -50,6 +50,7 @@ var performanceMarkPresent =
     global.performance && typeof global.performance.mark === "function";
 var setImmediatePresent =
     global.setImmediate && typeof global.setImmediate === "function";
+var utilPromisify = global.process && require("util").promisify;
 
 describe("issue #59", function() {
     var context = {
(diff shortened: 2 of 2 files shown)


src/lolex-src.js
```diff
@@ -21,6 +21,7 @@ function withGlobal(_global) {
         hrtimePresent && typeof _global.process.hrtime.bigint === "function";
     var nextTickPresent =
         _global.process && typeof _global.process.nextTick === "function";
+    var utilPromisify = _global.process && require("util").promisify;
```
SimenB: `_global.process` as a "I am in node" check?

src/lolex-src.js
```diff
@@ -743,6 +744,21 @@ function withGlobal(_global) {
                 delay: timeout
             });
         };
+        if (typeof _global.Promise !== "undefined" && utilPromisify) {
+            clock.setTimeout[
+                utilPromisify.custom
```
fatso83: A small comment/[link](https://nodejs.org/dist/latest-v10.x/docs/api/util.html#util_util_promisify_custom) explaining what `utilPromisify.custom` is could be useful for people reviewing the code ... ;-) 

```suggestion
                // Symbol used for custom overrides. See https://nodejs.org/dist/latest-v10.x/docs/api/util.html#util_util_promisify_custom
                utilPromisify.custom
```
~If this is indeed a `Symbol`, it would be useful to know when it was introduced wrt which Node versions we support, as this might be breaking (for some reason).~
We support Node LTS version 10 and 12 at the moment, and is this was introduced in Node 8, there are no breaking changes :)

src/lolex-src.js
fatso83: Just be sure we are targetting the right `global` here ... Should not this strictly be `globalObject` (as that's where the promisify function will be required from in the runtime)? It's not certain that people will pass a custom global with all of the props of the runtime.

jeysal: > `_global.process` as a "I am in node" check?

Yeah this seems to be one of the popular ways
https://stackoverflow.com/questions/4224606/how-to-check-whether-a-script-is-running-under-node-js

> Just be sure we are targetting the right `global` here ... Should not this strictly be `globalObject` (as that's where the promisify function will be required from in the runtime)? It's not certain that people will pass a custom global with all of the props of the runtime.

TBH I don't think I have the necessary knowledge of what the use cases for non-`defaultImplementation`s / the `withGlobal` function are to judge whether there will be cases where the provided `_global` looks like a Node environment, but our actual global scope does not have `require`. However FWIW I think it would be unintuitive to have vastly different conditions for applying e.g. the `nextTick` and `util.promisify` special handling, so that's why I'd prefer this.

SimenB: `withGlobal` was added for jest support so we can pass in the jsdom global.

jeysal: So then it'll apply to the `jest-environment-jsdom` case anyway because it has `process`.
Imagining that I pass in e.g. a DOM global that does *not* have Node things like `process`, I would probably want it to not apply the promisify change (even though it could, using the outside `require`), just like it doesn't apply the `nextTick` change, I think?

SimenB: I think inspecting `_global` is correct - if the global used to instantiate `lolex` pretends to be from node (by having a `process` property) we should assume we can `require('util');` safely and go with that.

src/lolex-src.js
SimenB: Lolex supports node 8 (please only drop it in a v6 release if you want to, even though it's EOL) https://github.com/sinonjs/lolex/blob/0e14fcac17e0cb49d5f7dd58227bec78b18703b3/.travis.yml#L5-L9

As you say this features is supported in Node 8 though, so no breaking change regardless

jeysal:
Anything more to do here? @fatso83 :)
@SimenB not sure how important the storing `global.Promise` thing is. Since it's not directly related to this, I'd do a separate PR for it, or we can go without it and see if someone actually runs into an issue? Your call

SimenB:
>Anything more to do here?

You haven't addressed "Should do `setImmediate` at the same time".

> we can go without it and see if someone actually runs into an issue?

It _has_ hit people in Jest, which is why we have tests for checking that we still work even if people swap out globals (e.g. the whole reason https://github.com/facebook/jest/blob/master/scripts/babel-plugin-jest-native-globals.js exists. That transform only guards Jest's own source code though, not other libraries). I'm fine with doing it separately, it's nothing to block on

jeysal:
> You haven't addressed "Should do `setImmediate` at the same time".

Right, always miss the comments in the reviews themselves :sweat_smile: pushed :+1: 

> It has hit people in Jest

I know, it was more of a "are the people changing globals going to be the same people using lolex fake timers" ;D

jeysal:
(Build error also happens on master, `test` passes for me locally)

mroderick:
@jeysal I can't see why this branch would fail to build on Circle. Would you mind rebasing it, to see if that might solve the build error in Circle?

jeysal:
done

jeysal:
worked 🎉

mroderick:
@fatso83 are you happy with the changes made? Is there anything left to do?

jeysal:
(I've put @simenb's suggestion about capturing `global.Promise` on my to-do list for a follow up, but this one is complete I think)

mroderick:
`@sinonjs/fake-timers@6.0.1`

joshua-barnett:
By the way this doesn't work for `setTimeout (global)` because the `setTimeout[util.promisify.custom]` custom promisified function is not set correctly there.

I assume that's because this for loop is intent on copying the properties on `clock.setTimeout` to `setTimeout (global)`.

However, the `clock.setTimeout[util.promisify.custom]` is not copied.

PR #300, opened 2020-02-01
Move to circle ci

mroderick:
This PR moves the CI tests from TravisCI to CircleCI, ~~and also drops Node 8 while we're at it.~~


SimenB:
If the `lolex` package will be deprecated on npm, please don't drop node 8 as Jest won't be able to upgrade.

---

Also, why Circle CI instead if GitHub Actions?

mroderick:
> If the `lolex` package will be deprecated on npm, please don't drop node 8 as Jest won't be able to upgrade.

I guess it's a good thing that you discovered this PR then :)

Do you know of any schedule for dropping Node 8 support in Jest?

mroderick:
> Also, why Circle CI instead if GitHub Actions? (you also need to update the badge in the readme)

We're using CircleCI in all the other Sinon projects.

I am not particularly tied to using that, but don't have unlimited time for unpaid work. Learning and migrating to GitHub actions will take time.

mantoni:
I have migrated projects to GitHub actions recently and I'm happy to help. The Sinon project already uses GitHub actions and AFAICT the only bit that is missing is running the test suite on SauceLabs.

SimenB:
https://github.com/facebook/jest/blob/v25.1.0/packages/jest-fake-timers/package.json#L17

Schedule is "next major", but that major is probably a few months out.

Unless you want something from node 10, any reason to drop it beyond it being EOL (which I'm wholly for, it's the combination with deprecating the old package I'm against)

mroderick:
Any significant changes should probably include automating publishing of new releases, as that is sensitive to user error and availability of maintainers with rights and confidence in publishing to npm.

mroderick:
> https://github.com/facebook/jest/blob/v25.1.0/packages/jest-fake-timers/package.json#L17

Thanks!

> Schedule is "next major", but that major is probably a few months out.

I don't mind holding off on dropping Node 8 for a few months for just `lolex`.

> Unless you want something from node 10, any reason to drop it beyond it being EOL (which I'm wholly for, it's the combination with deprecating the old package I'm against)

Do you think it's acceptable to rename the package and deprecate the old one. And then in some months release a new major version that drops Node 8 support?

mantoni:
Regarding deprecation: I think once we've completed the renaming process of the project, we should cut a new major release and deprecate that new major version only. This way, a fresh `npm install lolex` shows the deprecation message. People depending on any of the current majors shouldn't start seening deprecation warnings for no good reason.

SimenB:
@mantoni that sounds like a great idea !👍 Best approach

mroderick:
I've updated this PR to re-add testing in Node 8.

mroderick:
I am happy for us to use @mantoni's recommendation for deprecation after renaming, however that's separate from the work in this PR. Seperate PRs will follow.

PR #316, opened 2020-03-11
fix incorrect paths in README

bertho-zero:


mantoni:
Nice! Thank you for contributing ❤️

Issue #276, opened 2019-11-20
lolex.install({target: otherContext}) injects properties from original global


fatso83:
* Lolex version : 5.1.1
* Environment   : Chrome

**What did you expect to happen?**
That `lolex.install({target: myContext, ...otherProps})` would basically work like `lolex.withGlobal(target).install({...otherProps})`. 

**What actually happens**
If your environment supports `performance` and you do `lolex.install({target: { toFake: ['setTimeout'] } })` it still fakes and injects `performance` into the returned clock. 

```
assert.isUndefined(clock.performance);
```

Will therefore fail. This happens because the checks on whether to add the prop references the prop that was set using the default lolex instance.

**How to reproduce**
```
function noop(){}
global.performance = {now:function(){}};
var lolex = require('./src/lolex-src');
var clock2 = lolex.install({target: {Date: noop}});
assert.isUndefined(clock2.performance); // fails
```

fatso83:
An easy fix would be to either disallow the `target` prop and explicitly say `"Update your code to use `lolex.withGlobal(target).install(config)"` or just do it under the covers in the install method:
```javascript
if(config.target) {
   var target = config.target;
   delete config.target;
   return withGlobal(target)(config);
}
```

SimenB:
What if I want to install fake versions of timers from one thing into some other object? Is that a valid use case?

fatso83:
@SimenB I can't really come up with any reason to do that ... but I guess someone _might_ want that ... for some unknown reason?

```javascript
const myTarget = { Date: sinon.fake() };
lolex.withGlobal(global.jsDomWindow).install({target: myTarget});
```

That would install fake `performance` timers into `myTarget`, so you could tamper with that object without worrying about messing too much with `global.jsDomWindow`. Not sure who would need such a use case, but we _can_ keep it. Then we would need to document it, with a hint to why you would use it.

SimenB:
I'm down with throwing on `target` usage and directing to use the factory function for that use case. It works with any use case I can think of at least

fatso83:
#318

SimenB:
This is fixed now, right?

PR #318, opened 2020-03-23
Remove config.target

fatso83:
Removes `config.target`, simplifying the API by
only making it possible to choose alternate targets
using `withGlobal(target)`.

See discussion in #276 and sinonjs/sinon#2156

### DIFF
--- a/integration-test/fake-clock-integration-test.js
+++ b/integration-test/fake-clock-integration-test.js
@@ -37,7 +37,7 @@ describe("withGlobal", function() {
     });
 
     it("should support basic setTimeout", function() {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -75,7 +75,7 @@
   "module": "./fake-timers.js",
   "main": "./src/fake-timers-src.js",
   "dependencies": {
-    "@sinonjs/commons": "^1.7.0"
+    "@sinonjs/commons": "1.7.1"
```
SimenB: ?

fatso83: Unintentional. me fix

PR #323, opened 2020-05-11
fix: fake all supported timers by default

SimenB:
It's very weird to me that this method is excluded from default mocking. I've read #126 and to me the solution seems to be for users to specify a `toFake` themselves, not force other people to specify "yes, when I say fake timers I do mean _all_ timers".

Note that this should be considered a breaking change.

Whenever we start supporting some _new_ timer, it should by default not be mocked, but I think when we later make a new major release it should be mocked by default again.


src/fake-timers-src.js
```diff
@@ -1262,10 +1262,7 @@ function withGlobal(_global) {
         clock.methods = config.toFake || [];
 
         if (clock.methods.length === 0) {
-            // do not fake nextTick by default - GitHub#126
-            clock.methods = keys(timers).filter(function(key) {
-                return key !== "nextTick" && key !== "queueMicrotask";
```
SimenB, line 1267: seems docs were lying 🤷

src/fake-timers-src.js
```diff
@@ -1259,14 +1259,7 @@ function withGlobal(_global) {
...
 
-        clock.methods = config.toFake || [];
-
-        if (clock.methods.length === 0) {
-            // do not fake nextTick by default - GitHub#126
-            clock.methods = keys(timers).filter(function(key) {
-                return key !== "nextTick" && key !== "queueMicrotask";
-            });
-        }
+        clock.methods = config.toFake || keys(timers);
```
SimenB, line 1262: another technically breaking change I guess - if people passed `toFake: []` before, we faked everything. Seems like a bug fix to me to respect it, but maybe not 😀 If _not_, should just revert the last commit

benjamingr:
So here's the thing: promise libraries and other libraries use nextTick and microticks for scheduling so this is a pretty big breaking change.

In principle I agree that we should mock this by default.

test/fake-timers-test.js
```diff
@@ -4504,19 +4504,6 @@ describe("FakeTimers", function() {
             assert(called);
             clock.uninstall();
         });
-
-        it("does not install by default - GitHub#126", function(done) {
```
benjamingr, line 4508: Would it be possible to change this test to assert nextTick is mocked if the behavior changes?

src/fake-timers-src.js
benjamingr: Yeah I think this and the other one should be a major version if we land them.

SimenB:
> So here's the thing: promise libraries and other libraries use nextTick and microticks for scheduling so this is a pretty big breaking change.

Good point, should probably call it out in the readme.

> In principle I agree that we should mock this by default.

👍

test/fake-timers-test.js
SimenB: changed the test above, I think that's fine?

SimenB:
@benjamingr added a section to the readme, feedback appreciated 😀

src/fake-timers-src.js
SimenB: Can open a separate PR for it?

benjamingr:
@SimenB I am +1 on this change in principle but I know it will break pretty much every fake-timers install "in the wild" that uses the current API.

SimenB:
Should work if people use the builtin `Promise` and `async-await`, no? If it breaks builtin `Promise` global I'd consider that a bug in node.

My main issue is that to enable "fake everything" I have to enumerate all timers from the outside, which is not obvious I think. Maybe we could rather add a `fakeNextTick` explicit option (defaulting to `false` I guess) instead of piggy-backing `toFake`?

benjamingr:
> Should work if people use the builtin Promise and async-await, no? If it breaks builtin Promise global I'd consider that a bug in node.

It doesn't break the built in promises but mostly because we don't mock its timings at the moment (we probably should with async_hooks). In practice this means that it's a lot harder than it has to to test promises with fake-timers (I recall a lot of discussions about this + async versions of timers somewhere?)

I don't mind seeing what breaks with mocking nextTick on some large codebases I maintain - though I assume I know the answer is "every place that relies on nextTick executing and doesn't .tick or runMicrotasks before that :]

SimenB:
In a work project where I migrated to use Jest's fake timers now that Jest ships with Lolex it broke `node-fetch` (and/or integration with `nock`, not entirely sure)

mroderick:
Stale bot?!

fatso83:
This makes my head hurt and you guys know best anyhow. Not sure why it has stalled, though?

fatso83 (approved):
To me, this looks good. @SimenB anything you want addressed (apart from the conflicts)?

fatso83:
B,B,B,B,bump? It's not super problematic to release a major version, as long as you also document how to deal with the fallout. Like using node-fetch as an example case, show what needs to be done to fix the test cases. @SimenB

src/fake-timers-src.js
fatso83: > Can open a separate PR for it?

Yes

test/fake-timers-test.js
fatso83: > changed the test above, I think that's fine?

Yes, this is fine. It's implicitly true, as the code won't work unless `nextTick` is synchronous.

fatso83 (approved):

mroderick:
This needs a rebase to resolve the conflicts.

@SimenB do you have time to take this across the finish line? If not, would you be ok with someone else doing it?

fatso83:
**I want to finish this (and do #490) and release a new breaking major. Soonish** 

> I don't mind seeing what breaks with mocking nextTick on some large codebases I maintain

There's been 4 years. I guess the number of places that rely on non-native Promises are a lot fewer and further between. Maybe it's not that bad? I think we would just need to add something to the changelog.

fatso83:
Since I closed #490 and I haven't heard anything in months, I think I'll just pick this up and release a new breaking major.

fatso83:
Weird how much you can get done in a weekend without kids 😄

fatso83:
I think this needs to be reverted. It seems to trip people up over and over again, ref https://github.com/sinonjs/sinon/issues/2620#issuecomment-2461063710, and so it makes for poor developer experience. `nextTick` and `queueMicroTask` is far less common to target than `setTimeout` and similar, but it's the latter crowd that now needs to supply a list of timers, instead of the more exotic crowd.

I'd rather just change the docs to say "fakes _most_ timers by default, but you can also choose to stub all or a subset, if you so please" and then expose some constants they could use (`DEFAULT_TIMERS`,`ES_STANDARD_TIMERS`, `ALL_TIMERS`), a la:
```javascript
import 
fakeTimers.install({toFake: fakeTimers.ALL_TIMERS})
```

Issue #230, opened 2019-02-05
Provide more context on infinites recursion


SimenB:
> We understand you have a problem and are in a hurry, but please provide us with some info to make it much more likely for your issue to be understood, worked on and resolved quickly.

* Lolex version : 3.0
* Environment   : N/A
* Example URL   : N/A
* Other libraries you are using: N/A

**What did you expect to happen?**

See https://github.com/facebook/jest/issues/2893 for details.

Tl;dr: Right now Lolex just throws an error if it thinks we're in an infinite loop of timers (timers scheduling new timers). This comes from here: https://github.com/sinonjs/lolex/blob/29b4ecc8bd186584ab4139341f3128a4349463ba/lolex.js#L210

It would be nice if the message thrown included a stack trace back to the code actually scheduling timers

**What actually happens**

An error without context

**How to reproduce**
https://runkit.com/embed/jqyax6pkc5bq

benjamingr:
We can add the stack to clocks when they are captured and show it on the clock when this is thrown - though this will be _really_ expensive CPU wise.

SimenB:
What about, after reaching `loopLimit`, you schedule 10 more or so, and capture just their stacks, then throw? So in the normal case there is no change, but you loop a few extra times (maybe even `loopLimit - 10` to not really affect anything observable) in the case where things are gonna throw anyways?

benjamingr:
So:

 - If I'm inside lolex and I'm at `loopLimit - 10`
 - Then start capturing stack traces when setTimeout/setInterval etc are called
 - If the loop limit is reached before another timer is hit, fail without a stack
 - Otherwise fail with the stack any of those timers.

Sounds useful/good to me and like it should solve the 95% case.

SimenB:
Yeah, that sounds reasonable to me! Potentially start with just a single stack instead of 10, not sure? Changing that threshold is simple once the logic is in place though, so not something we need to spend time on deciding now

SimenB:
>* If the loop limit is reached before another timer is hit, fail without a stack

Wait, no - not this. If `loopLimit` is 100, we don't want to fail if we get to 95 timers before bailing

alistairjcbrown:
Put up PR #325 to add stack trace output for the next excess job to be run - let me know if you think that's the right direction for solving this issue

PR #325, opened 2020-05-17
Add stack trace to code recursively scheduling timers

alistairjcbrown:

Fix issue #230 by using stack information for jobs and timers to provide more context when an infinite loop error is thrown.

#### Solution  - optional

This solution works by adding a `stack` property to jobs and tasks. This data can then be used, if an infinite loop is detected, to set the stack of the thrown `infiniteLoopError`.

##### Example stack trace from test run
```
Error: Aborting after running 5 timers, assuming an infinite loop!
Timeout - recursiveCreateTimerTimeout
      at recursiveCreateTimer (test/fake-timers-test.js:4668)
      at recursiveCreateTimerTimeout (test/fake-timers-test.js:4669)
      at callTimer (src/fake-timers-src.js:441)
      at Object.next (src/fake-timers-src.js:1039)
      at src/fake-timers-src.js:1128
```

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -67,6 +67,18 @@ function withGlobal(_global) {
     var NativeDate = _global.Date;
     var uniqueTimerId = 1;
 
+    var isNearInfiniteLimit = false;
+
+    function checkIsNearInfiniteLimit(clock, i) {
+        if (clock.loopLimit && i === clock.loopLimit - 1) {
+            isNearInfiniteLimit = true;
+        }
+    }
+
+    function resetIsNearInfiniteLimit() {
+        isNearInfiniteLimit = false;
+    }
+
     function isNumberFinite(num) {
         if (Number.isFinite) {
             return Number.isFinite(num);
@@ -275,10 +287,13 @@ function withGlobal(_global) {
         for (var i = 0; i < clock.jobs.length; i++) {
             var job = clock.jobs[i];
             job.func.apply(null, job.args);
+
+            checkIsNearInfiniteLimit(clock, i);
             if (clock.loopLimit && i > clock.loopLimit) {
                 throw getInfiniteLoopError(clock.loopLimit, job);
             }
         }
+        resetIsNearInfiniteLimit();
         clock.jobs = [];
     }
 
@@ -287,7 +302,9 @@ function withGlobal(_global) {
             throw new Error("Callback must be provided to timer calls");
         }
 
-        timer.error = new Error();
+        if (isNearInfiniteLimit) {
+            timer.error = new Error();
+        }
 
         timer.type = timer.immediate ? "Immediate" : "Timeout";
 
@@ -795,7 +812,7 @@ function withGlobal(_global) {
             return enqueueJob(clock, {
                 func: func,
                 args: Array.prototype.slice.call(arguments, 1),
-                error: new Error()
+                error: isNearInfiniteLimit ? new Error() : null
             });
         };
 
@@ -1090,15 +1107,18 @@ function withGlobal(_global) {
             runJobs(clock);
             for (i = 0; i < clock.loopLimit; i++) {
                 if (!clock.timers) {
+                    resetIsNearInfiniteLimit();
                     return clock.now;
                 }
 
                 numTimers = keys(clock.timers).length;
                 if (numTimers === 0) {
+                    resetIsNearInfiniteLimit();
                     return clock.now;
                 }
 
                 clock.next();
+                checkIsNearInfiniteLimit(clock, i);
             }
 
             var excessJob = firstTimer(clock);
@@ -1119,6 +1139,7 @@ function withGlobal(_global) {
                                 var numTimers;
                                 if (i < clock.loopLimit) {
                                     if (!clock.timers) {
+                                        resetIsNearInfiniteLimit();
                                         resolve(clock.now);
                                         return;
                                     }
@@ -1126,6 +1147,7 @@ function withGlobal(_global) {
                                     numTimers = Object.keys(clock.timers)
                                         .length;
                                     if (numTimers === 0) {
+                                        resetIsNearInfiniteLimit();
                                         resolve(clock.now);
                                         return;
                                     }
@@ -1135,6 +1157,7 @@ function withGlobal(_global) {
                                     i++;
 
                                     doRun();
+                                    checkIsNearInfiniteLimit(clock, i);
                                     return;
                                 }
 
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -143,6 +143,30 @@ function withGlobal(_global) {
         throw new TypeError("now should be milliseconds since UNIX epoch");
     }
 
+    function getInfiniteLoopError(loopLimit, job) {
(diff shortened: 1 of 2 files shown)


src/fake-timers-src.js
```diff
@@ -767,7 +786,11 @@ function withGlobal(_global) {
         clock.nextTick = function nextTick(func) {
             return enqueueJob(clock, {
                 func: func,
-                args: Array.prototype.slice.call(arguments, 1)
+                args: Array.prototype.slice.call(arguments, 1),
+                stack: new Error().stack
```
SimenB, line 790: `.stack` is potentially somewhat expensive, could we access it lazily in the error case?

SimenB:
Could we just keep track of the last scheduled timer instead of all scheduled timers? Not sure how much that matters

src/fake-timers-src.js
alistairjcbrown: @SimenB I'm not sure how we can get the right stack trace unless we generate it when the job is created. Just to see the difference, I added these properties locally (and the same equivalent in `nextTick`):
```js
        timer.fullStack = new Error().stack;
        timer.getStackLazy = function() { return new Error().stack };
```

And then logged them out within `getInfiniteLoopError`:
```js
        console.log(job.fullStack);
        console.log('---');
        console.log(job.getStackLazy());
```

The stack trace generated lazily provides the running stack trace (i.e. `runMicrotasks` or `runAll` calls) rather than the stack trace of the job itself.

❓ Is there a different way we can do it?

<details>
    <summary>Stack trace comparison - first stack is generated when adding the job, second is generated lazily when an infinite loop is detected</summary>



#### Running microtasks
```
Error
      at Object.nextTick (src/fake-timers-src.js:803)
      at Object.queueMicrotask (src/fake-timers-src.js:809)
      at recursiveQueueMicroTask (test/fake-timers-test.js:4639)
      at runJobs (src/fake-timers-src.js:277)
      at Object.runMicrotasks (src/fake-timers-src.js:880)
      at Context.<anonymous> (test/fake-timers-test.js:4648)
---
Error
      at Object.getStackLazy (src/fake-timers-src.js:804)
      at getInfiniteLoopError (src/fake-timers-src.js:156)
      at runJobs (src/fake-timers-src.js:279)
      at Object.runMicrotasks (src/fake-timers-src.js:880)
      at Context.<anonymous> (test/fake-timers-test.js:4648)
```

#### Running all timers
```
Error
      at addTimer (src/fake-timers-src.js:295)
      at Object.setTimeout (src/fake-timers-src.js:769)
      at target.<computed> (src/fake-timers-src.js:595)
      at recursiveCreateTimer (test/fake-timers-test.js:4668)
      at recursiveCreateTimerTimeout (test/fake-timers-test.js:4669)
      at callTimer (src/fake-timers-src.js:450)
      at Object.next (src/fake-timers-src.js:1050)
      at src/fake-timers-src.js:1139
---
Error
      at Object.timer.getStackLazy (src/fake-timers-src.js:296)
      at getInfiniteLoopError (src/fake-timers-src.js:156)
      at src/fake-timers-src.js:1149
```

</details>

SimenB: I meant to still do `new Error()` here, but only access `stack` if we actually wanna throw the error

SimenB:
Another though, this is mostly meant to catch timers added while running timers. Could we add some flag internally that's set to `true` whenever we run timers, and only collect these at that time?

alistairjcbrown:
@SimenB Thanks for the feedback!
I've updated to set an error object on the job instead of getting the error stack. The second commit adds in a `isNearInfiniteLimit` flag which is used to decide whether to attach an error object to the job. Jobs from the job before we hit the limit will have an error property populated, which is then used if we need to create an InfiniteLoopError. What do you think?

fatso83:
no time (parental leave) at the moment to look at PRs, sorry.

src/fake-timers-src.js
```diff
@@ -143,6 +155,30 @@ function withGlobal(_global) {
...
     }
 
+    function getInfiniteLoopError(loopLimit, job) {
+        var infiniteLoopError = new Error(
+            "Aborting after running " +
+                loopLimit +
+                " timers, assuming an infinite loop!"
+        );
+
+        var sliceSize = job.type ? 4 : 3;
```
benjamingr, line 165: It would be ideal to calculate this by looking for ourselves in the stack trace rather than setting prefixed sizes. This can fail based on adding new jobs.

src/fake-timers-src.js
```diff
@@ -143,6 +155,30 @@ function withGlobal(_global) {
...
+
+        infiniteLoopError.stack =
```
benjamingr, line 167: Some environments don't have a writeable stack property - so it would be better to wrap this in a try/catch and to use defineProperty

benjamingr: In particular we had this issue several times in bluebird which also mutates stacks

benjamingr:
Also:

>  What do you think?

The actual implementation logic lgtm

alistairjcbrown:
@benjamingr Awesome, thanks for the feedback! I'll take a run through it tomorrow night 👍

alistairjcbrown:
@benjamingr 

> Dealing with a non-writeable stack property

Using `defineProperty` wrapped in `try`/`catch` as suggested - TIL about non-writeable stacks 👍 

> calculating the slice size automatically

Doing this by running through the stack line by line, checking the line matches `at target.X ` (where `X` may be `(anonymous function)` or `<computed>`) or `at Object.Y` (where `Y` is a clock method).

> Tests that check a larger chunk of the stack (so we're sure this doesn't break)

Apologies, I wasn't 100% sure on what this was asking, so I've added more tests to cover `requestAnimationFrame`, `setImmediate`, `setInterval`, `requestIdleCallback`,  `nextTick`.

If this was referring to checking more of the stack trace itself, the tests are checking that we're getting the expected method at the top of the stack - but we can't check the filename as it'll be `__puppeteer_evaluation_script__` on headless. Do you want more of the stack trace checked?

benjamingr:
Hey, I wanted to merge and it looks like there is a merge conflict - can you please take a look?

alistairjcbrown:
@benjamingr Rebased off `master` and conflict resolved 👍

benjamingr:
Thanks!

fatso83:
This broke the builds on Safari, IE and Firefox. It needs some patching before we can release it as part of a new release.
See https://app.circleci.com/pipelines/github/sinonjs/fake-timers/117/workflows/625d68b6-ed2c-4d04-a450-c22a240599dd/jobs/670

The "cloud" tests are only run on `master`, as PR runs can expose the Sauce Labs credentials.

fatso83:
@benjamingr Do you think you can get this in again, after fixing the issues that broke the build? Was some good work done here by @alistairjcbrown. Shame to let it go to waste.

benjamingr:
Yeah I agree - I just didn't have a ton of free time recently :/ 

Maybe @itayperry wants to pick up readding this? (I can guide, I just have little free time to dedicate to this myself)

itayperry:
I would be happy to take part in this @benjamingr 🎉🪴 I could, of course, use some guidance and tips. I hope I'll be able to help 💭

Issue #319, opened 2020-03-23
Remove use of `eval`


fatso83:
* FakeTimers version : 6.0
* Environment   : Rollup
* Example URL   : https://github.com/sinonjs/nise/issues/110

**What did you expect to happen?**
No warnings/errors

**What actually happens**
Rollup complaints about the use of `eval`.

**How to reproduce**
[See original bug report.](https://github.com/sinonjs/nise/issues/110)

Judging by the comments in the original issue in nise, this issue seems to be caused by needing to support some ancient browsers. Fixing it should thus be easy, but requires a new major version.

mantoni:
We had a discussion about this before, I recall, but can't find it. We should throw a meaningful error if a string is passed as the first argument to `setTimeout` or `setInterval`.

I suggest to throw the same error as Node:

```
$ node -p 'setTimeout("void()",1)'

TypeError [ERR_INVALID_CALLBACK]: Callback must be a function. Received 'void()'
```

itayperry:
Hey everyone, I'm taking a look at this. I'll try to fix the issue :)

itayperry:
Hey there, I removed the `eval()` function, it is no longer in use. It would be great if you take a look at the code and check that it doesn't break anything. There was a test for the `eval()` and I changed it to make sure that it fails and **throws an exception** :)

benjamingr:
I just did a sanity check and it looks like this feature is still supported in all major browsers.

In https://github.com/sinonjs/nise/issues/110 and specifically https://github.com/sinonjs/nise/issues/110#issuecomment-550090232 @mantoni saw that it didn't work in Chrome - but that's only because the page he was testing in had CSP (for example GitHub.com) - sites like example.com will happily execute code like `setTimeout("console.log('hello world')")`.

Chrome:

![image](https://user-images.githubusercontent.com/1315533/82688668-ef5f9180-9c61-11ea-942d-c7202bc9a81a.png)

In addition checking on the grid, it looks like this feature still works on IE11:

![image](https://user-images.githubusercontent.com/1315533/82688619-db1b9480-9c61-11ea-8b23-3419d704796e.png)

MSHtml Edge:

![image](https://user-images.githubusercontent.com/1315533/82688740-0a320600-9c62-11ea-9191-3872261b896b.png)

Firefox:

![image](https://user-images.githubusercontent.com/1315533/82688992-79a7f580-9c62-11ea-9ede-6c62adb50e9a.png)

Safari:

[...]

PR #327, opened 2020-05-22
Remove use of `eval`

itayperry:
#### Purpose 
Removing the use of `eval()` as there are 'Rollup complaints about the use of eval.'
According to MDN: 

> 'Executing JavaScript from a string is an enormous security risk'

Issue #319

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -421,8 +421,12 @@ function withGlobal(_global) {
         if (typeof timer.func === "function") {
             timer.func.apply(null, timer.args);
         } else {
-            /* eslint no-eval: "off" */
-            eval(timer.func);
+            throw new Error(
+                "TypeError [ERR_INVALID_CALLBACK]: Callback must be a function. Received " +
+                    timer.func +
+                    " of type " +
+                    typeof timer.func
+            );
         }
     }
 
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -421,8 +421,8 @@ function withGlobal(_global) {
         if (typeof timer.func === "function") {
             timer.func.apply(null, timer.args);
         } else {
-            throw new Error(
-                "TypeError [ERR_INVALID_CALLBACK]: Callback must be a function. Received " +
+            throw new TypeError(
+                "[ERR_INVALID_CALLBACK]: Callback must be a function. Received " +
                     timer.func +
                     " of type " +
                     typeof timer.func
--- a/test/fake-timers-test.js
+++ b/test/fake-timers-test.js
@@ -330,7 +330,7 @@ describe("FakeTimers", function() {
                 }.bind(this),
                 {
                     message:
-                        "TypeError [ERR_INVALID_CALLBACK]: Callback must be a function. Received " +
+                        "[ERR_INVALID_CALLBACK]: Callback must be a function. Received " +
                         notTypeofFunction +
                         " of type " +
                         typeof notTypeofFunction
--- a/test/fake-timers-test.js
+++ b/test/fake-timers-test.js
@@ -320,11 +320,22 @@ describe("FakeTimers", function() {
             assert(FakeTimers.evalCalled);
         });
 
-        it("evals non-function callbacks", function() {
-            this.clock.setTimeout("FakeTimers.evalCalled = true", 10);
-            this.clock.tick(10);
+        it("does not accept non-function callbacks", function() {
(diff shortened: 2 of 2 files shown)


src/fake-timers-src.js
```diff
@@ -421,8 +421,12 @@ function withGlobal(_global) {
         if (typeof timer.func === "function") {
             timer.func.apply(null, timer.args);
         } else {
-            /* eslint no-eval: "off" */
-            eval(timer.func);
+            throw new Error(
+                "TypeError [ERR_INVALID_CALLBACK]: Callback must be a function. Received " +
```
mantoni, line 425: Can you change the `Error` to be a `TypeError` and remove the "TypeError" prefix from the message? This would currently serialize as `Error: TypeError …`.

It would also help with readability of the message if the actual received value would be quoted.

test/fake-timers-test.js
```diff
@@ -320,11 +320,22 @@ describe("FakeTimers", function() {
             assert(FakeTimers.evalCalled);
         });
 
-        it("evals non-function callbacks", function() {
-            this.clock.setTimeout("FakeTimers.evalCalled = true", 10);
-            this.clock.tick(10);
+        it("does not accept non-function callbacks", function() {
+            var notTypeofFunction = "FakeTimers.evalCalled = true";
+            this.clock.setTimeout(notTypeofFunction, 10);
```
mantoni, line 325: This is where the exception should be thrown, as this is the place where the programmer made the mistake.

src/fake-timers-src.js
itayperry: Delete the 'TypeError' and let it start with [ERR_INVALID_CALLBACK]?

test/fake-timers-test.js
itayperry: The use of `eval()` was present in the `callTimer()` function and not in the `addTimer()` function. You are basically right. I'm afraid that changing the `addTimer()` would cause many tests to fails and it might be a radical change. But I'll look into it :)

src/fake-timers-src.js
mantoni: Yes. And throw a `TypeError` instead of an `Error`.

test/fake-timers-test.js
mantoni: Don't worry. It shouldn't break any existing tests, except the ones that should, because a string is passed to `setTimeout`.

src/fake-timers-src.js
itayperry: Ohhh ok! Now I get it :) I never used a different type of error before, thanks for teaching me something new!

test/fake-timers-test.js
itayperry: There is a function called `addTimer()` that actually creates all these timers:

Here it is:
https://github.com/sinonjs/fake-timers/blob/89e9e9b58c12fdf3301067dee6282ca829c48c56/src/fake-timers-src.js#L265

If I add this code to it:
```
if (typeof timer.func !== "function") {
            throw new TypeError(
                "[ERR_INVALID_CALLBACK]: Callback must be a function."
            );
}
```
It will break these 4 tests: 

https://github.com/sinonjs/fake-timers/blob/89e9e9b58c12fdf3301067dee6282ca829c48c56/test/fake-timers-test.js#L275

https://github.com/sinonjs/fake-timers/blob/89e9e9b58c12fdf3301067dee6282ca829c48c56/test/fake-timers-test.js#L285

https://github.com/sinonjs/fake-timers/blob/89e9e9b58c12fdf3301067dee6282ca829c48c56/test/fake-timers-test.js#L3097

https://github.com/sinonjs/fake-timers/blob/89e9e9b58c12fdf3301067dee6282ca829c48c56/test/fake-timers-test.js#L3087

mantoni: That is correct. These tests pass in a string instead of a function. However, they actually test something else. I guess it's reasonable to change those tests and pass in functions.

itayperry: @benjamingr What do you think?

benjamingr: @itayperry I'm not sure what I'm being asked here but everything Maximilian has been saying sounded like useful and reasonable feedback and I encourage you to read these tests by stepping through them with a debugger.

itayperry:
As decided in #319 `eval()` will not be entirely canceled, for further info read:  https://github.com/sinonjs/fake-timers/issues/319#issuecomment-632794814

PR #329, opened 2020-05-30
Fixing issue #319 - preventing Rollup eval error

itayperry:
#### Purpose
This is the first step (out of 3) in solving issue #319.
The use of `eval()` will eventually not be canceled, and this pull request tricks Rollup to not throw an error when `eval()` is used.


benjamingr:
Hey, thanks, can you please add a test that eval runs on global scope?

itayperry:
I don't really know how to test it, I would be to get a piece of advice. There are a few tests that check if the `eval()` works. One of them is: https://github.com/sinonjs/fake-timers/blob/89e9e9b58c12fdf3301067dee6282ca829c48c56/test/fake-timers-test.js#L323

benjamingr:
Define a local variable, make a setTimeout, assert that it is not defines in the timeout (by calling tick)

itayperry:
The function:
```
var eval2 = eval;
(function() {
    eval2(timer.func);
})();
```
gets a variable that will always be global to it.

I understand that if I do something like: 

```
var eval2 = eval;
(function() {
    var x = 5; 
    eval2('x') 
    eval2(timer.func);
})();
```

It will throw a **reference error** saying that `x is undefined`, but in the way that the function is built now there's no way (from the outside through `setTimeout`) to insert a variable that will be local in this function.

I've tried doing something like `setTimeout('var x = 5', 10)` but this triggers a **leak error**.
I'm sorry for sending so many questions, but I wanna solve this issue perfectly :)

fake-timers.js
```diff
@@ -1726,7 +1726,10 @@ function withGlobal(_global) {
             timer.func.apply(null, timer.args);
         } else {
             /* eslint no-eval: "off" */
-            eval(timer.func);
```
benjamingr, line 1729: We should probably make another PR to make sure the node behavior is more correct.

itayperry: So this PR is ok for now? 
And I need to find out how this package "knows" on which env it is, because I'm aware that on Node it needs to **not** be possible at all. Originally overall there should be 3 PRs.

benjamingr: Yes, this PR is ok, I will give other members a few days to look at it and if all looks good we'll land it

itayperry: By the way, thank you for helping all the time. Your help was, and is, needed! :)

benjamingr:
@fatso83 how is the baby situation? Any chance to get a release? There are some pending changes and I'm scared to release myself because I haven't released fake-timers yet.

Issue #304, opened 2020-02-18
Error: Cannot clear timer: timer created with setInterval() but cleared with clearTimeout()


slightlytwisted:
* FakeTimers version : latest release
* Environment   : NodeJS 12.13.0

fake-timers should not assert that calls to `clearTimer()` be passed only timers created by `createTimer()` and that calls to `clearInterval()` be passed only timers created by `createInterval()` because the [HTML Living Standard](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-setinterval) indicates that:

> Because clearTimeout() and clearInterval() clear entries from the same list, either method can be used to clear timers created by setTimeout() or setInterval().

The [assertion](https://github.com/sinonjs/fake-timers/blob/0beb3c3bd328570c6fbde0e32604ca005e2cc331/src/fake-timers-src.js#L457) in fake-timer should probably be relaxed.

fatso83:
Love a bug report with a WHATWG link to the reference ❤️

itayperry:
Hey everyone, I'm taking a look at this :)

itayperry:
A question :) This `clearTimer()` function appears in 2 different files, why is that? Should I change it in both files? I would love for some help. 

- First location: https://github.com/sinonjs/fake-timers/blob/7f8322effa73a94dc43896cdfb49d63e821b32d1/fake-timers.js#L1733
- Second location: https://github.com/sinonjs/fake-timers/blob/7f8322effa73a94dc43896cdfb49d63e821b32d1/src/fake-timers-src.js#L429

benjamingr:
@itayperry fake-timers is generated from fake-timers-src through bundling it: https://github.com/sinonjs/fake-timers/blob/7f8322effa73a94dc43896cdfb49d63e821b32d1/package.json#L24 - you should generally only need to edit fake-timers-src.

itayperry:
All ready and tested :) 
I hope this solves the issue properly.
#324

benjamingr:
Fixed in https://github.com/sinonjs/fake-timers/pull/324/commits/1f83b99639f57a3980847da7aaddcf3d11308b5d

NearHuscarl:
In the mean time, I've got my test working again by patching the code from the commit mentioned above using [`patch-package`](https://github.com/ds300/patch-package). Here is the instructions for anyone wants to do the same

[...]

PR #330, opened 2020-06-12
issue #304 second fix - clearTimeout clears setInterval and clearInterval clears setTimeout

itayperry:
## Purpose
This is a follow up to pull request #324 

> Allowing clearTimeout to clear setInterval and clearInterval to clear setTimeout,
as this is possible by HTML Living Standard.
Issue #304

After starting a simple Rollup project and using this package locally through `npm-link`  I found out that the problem wasn't solved..

Apparently, only one file is being tested and the same function [`clearTimer()`] exists in two different files. Therefore, I found out that only changing `fake-timers-src.js` was not enough to actually make the change, but only to pass the test. It seems that `fake-timers.js` isn't being tested.

For further info read: https://github.com/sinonjs/fake-timers/issues/304#issuecomment-629370417 and also https://github.com/sinonjs/fake-timers/issues/304#issuecomment-629640406

**Oh, and one more thing**: The file `fake-timers.js` changes entirely (as opposed to `fake-timers-src.js`) when you click save in `VSCode`, probably because of the auto-indent option. Could anyone please check it for me? I just opened it on an old sublime to not change gazillion lines :)

### DIFF
--- a/fake-timers.js
+++ b/fake-timers.js
@@ -1751,7 +1751,10 @@ function withGlobal(_global) {
         if (clock.timers.hasOwnProperty(id)) {
             // check that the ID matches a timer of the correct type
             var timer = clock.timers[id];
-            if (timer.type === ttype) {
+            if (timer.type === ttype ||
+                (timer.type === "Timeout" && ttype === "Interval") ||
+                (timer.type === "Interval" && ttype === "Timeout")
+            ) {
                 delete clock.timers[id];
             } else {
                 var clear =


benjamingr:
fake-timers.js is a generated file, you shouldn't need to change it manually. Here is where it is generated: https://github.com/sinonjs/fake-timers/blob/master/package.json#L24

This doesn't happen on every pull request - it happens when the version is bumped. https://github.com/sinonjs/fake-timers/blob/master/scripts/version.sh#L17

itayperry:
So in theory, if I do `npm run bundle` to check my changes locally, it would work? 
And I'll close this pull request :)

benjamingr:
Right

itayperry:
I learned something new, thank you :)

PR #332, opened 2020-07-01
Remove notes on v2.0 from the docs

Alexsey:

Version 2.0 was released 3 years ago so there is no need to add focus on the fact that some feature was added at that moment


Alexsey:
what can I do for this to be merged?

fatso83 (approved):
Seems like a decent update.

PR #340, opened 2020-10-22
Remove misleading comment

mroderick:
This repository doesn't use the `no-prototype-methods` rule that is
managed with `local-rules` plugin, so the comments make no sense.

ESLint should really throw an error here...

#### How to verify

1. Check out this branch
1. search the codebase for `local-rules`
1. Observe that there are no results

### DIFF
--- a/fake-timers.js
+++ b/fake-timers.js
@@ -105,7 +105,6 @@ module.exports = function every(obj, fn) {
     var pass = true;
 
     try {
-        /* eslint-disable-next-line local-rules/no-prototype-methods */
         obj.forEach(function() {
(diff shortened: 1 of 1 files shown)


benjamingr (approved):

SimenB:
>ESLint should really throw an error here...

We can either run eslint with [`--report-unused-disable-directives`](https://eslint.org/docs/user-guide/command-line-interface#-report-unused-disable-directives) or enable https://github.com/mysticatea/eslint-plugin-eslint-comments/blob/master/docs/rules/no-unused-disable.md

mantoni (approved):

PR #349, opened 2021-01-12
Copy all property descriptors to installed methods

### LINKED ISSUE #347: Support util.promisify on installed functions
* FakeTimers version : 6.0.1
* Environment   : Node v15.5.1

[`util.promisify`](https://nodejs.org/api/util.html#util_util_promisify_original) would work on `setTimeout` installed by FakeTimers as it does on the global `setTimeout`.

**What actually happens**

The Promise returned by the installed `setTimeout` is never resolved because the installed `setTimeout` does not have a [`promisify.custom`](https://nodejs.org/api/util.html#util_custom_promisified_functions) symbol property.

**How to reproduce**

```js
const { promisify } = require('util');
const FakeTimers = require("@sinonjs/fake-timers");

console.log('initial global.setTimeout', setTimeout[promisify.custom]);

const target = { Date };
FakeTimers.install({ target });
console.log('installed target.setTimeout', target.setTimeout[promisify.custom]);

FakeTimers.install();
console.log('installed global.setTimeout', setTimeout[promisify.custom]);
```
prints

    initial global.setTimeout [Function]
    installed target.setTimeout undefined
    installed global.setTimeout undefined

**Discussion**

This issue was noted by @joshua-barnett in https://github.com/sinonjs/fake-timers/pull/292#issuecomment-744443087.  I'm opening this issue to track it.

kevinoid:

Fix #347 by copying symbol own-properties to installed methods.

The `for...in` loop used to copy properties only copies string own-properties to installed methods, which omits [`util.promisify.custom`](https://nodejs.org/api/util.html#util_custom_promisified_functions) necessary for [`util.promisify`](https://nodejs.org/api/util.html#util_util_promisify_original) support on Node as added in #292.

#### Solution  - optional

This PR copies all own-property descriptors to get both `Symbol` own-properties and to copy any getters/setters that may be added in the future.

If there's a reason getters/setters should not be copied, let me know and I can add an `Object.assign` polyfill to [sinonjs/commons](https://github.com/sinonjs/commons) and use that instead.

Thanks for considering,
Kevin

Fixes: #347

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -535,7 +535,6 @@ function withGlobal(_global) {
     }
 
     function hijackMethod(target, method, clock) {
-        var prop;
         clock[method].hadOwnProperty = Object.prototype.hasOwnProperty.call(
             target,
             method
@@ -575,11 +574,10 @@ function withGlobal(_global) {
                 return clock[method].apply(clock, arguments);
             };
 
-            for (prop in clock[method]) {
-                if (clock[method].hasOwnProperty(prop)) {
-                    target[method][prop] = clock[method][prop];
-                }
-            }
+            Object.defineProperties(
+                target[method],
+                Object.getOwnPropertyDescriptors(clock[method])
+            );
         }
 
         target[method].clock = clock;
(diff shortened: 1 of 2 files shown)


benjamingr (approved):

fatso83:
I don't know why this has been dormant for five months, but I see @benjamingr approved it, so I guess it's good. Just merge it the next time, Benjamin, unless there's something specific you want a second pair of eyes on (in which case you just ping that someone) 🙂

PR #350, opened 2021-01-13
Update README section about browser use

mroderick:
In 7f56133373b240a397b00210c6b22c64d8ca8154 we removed the bundled
version, but failed to update instructions for use in browsers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,11 +22,7 @@ wait.
 npm install @sinonjs/fake-timers
 ```
 
-If you want to use `@sinonjs/fake-timers` in a browser you can use [the pre-built
-version](https://github.com/sinonjs/fake-timers/blob/master/fake-timers.js) available in the repo
-and the npm package. Using npm you only need to reference `./node_modules/@sinonjs/fake-timers/fake-timers.js` in your `<script>` tags.
-
-You are always free to [build it yourself](https://github.com/sinonjs/fake-timers/blob/53ea4d9b9e5bcff53cc7c9755dc9aa340368cf1c/package.json#L22), of course.
+If you want to use `@sinonjs/fake-timers` in a browser you can either build your own bundle or use [Skypack](https://www.skypack.dev).
 
 ## Usage
 


README.md
```diff
@@ -22,11 +22,7 @@ wait.
 npm install @sinonjs/fake-timers
 ```
 
-If you want to use `@sinonjs/fake-timers` in a browser you can use [the pre-built
-version](https://github.com/sinonjs/fake-timers/blob/master/fake-timers.js) available in the repo
-and the npm package. Using npm you only need to reference `./node_modules/@sinonjs/fake-timers/fake-timers.js` in your `<script>` tags.
-
-You are always free to [build it yourself](https://github.com/sinonjs/fake-timers/blob/53ea4d9b9e5bcff53cc7c9755dc9aa340368cf1c/package.json#L22), of course.
+If you want to use `@sinonjs/fake-timers` in a browser you can either build your own bundle or use [Skypack](https://www.skypack.dev).
```
fatso83, line 25: I guess this is something like https://wzrd.in, but does it understand the typescript config?

mroderick: > I guess this is something like https://wzrd.in

Except this one works with scoped packages (see https://github.com/browserify/wzrd.in/issues/104)

> but does it understand the typescript config?

I can't imagine why it would need to. Care to elaborate?

fatso83: I just did not understand what it was doing. I thought it did something like browserify, building the package, but now that I see that it's exclusively for evergreen browsers that support ESModules it makes more sense. I tried it right now, and it worked fine:

**index.html**
```javascript
<script type="module">
import * as lolex from 'https://cdn.skypack.dev/@sinonjs/fake-timers';
lolex.install();
console.log(new Date() ); // shows Jan 01 1970 :)
</script>
```

mroderick: Cheers

mroderick:
I chose to make a new release, so people viewing the README on non-GitHub websites and in the local filesystems would see the updated instructions for browser use.

This has been published to the npm registry as `@sinonjs/fake-timers@7.0.1`.

PR #353, opened 2021-01-17
Remove mention of `target` from readme

geigerzaehler:
#### Purpose (TL;DR)
The `target` option was removed in #318. We remove all mentions of this option from the readme and replace it with `withGlobal`.


fatso83:
Thanks!

PR #354, opened 2021-01-17
Make config properties optional in TypeScript defs

### LINKED ISSUE #352: Type definitions introduced in v7 are too strict
* FakeTimers version : 7.0.1
* Environment   : Node.js
* Example URL   : https://codesandbox.io/s/headless-lake-jq53p?file=/src/index.ts
* Other libraries you are using: TypeScript (4.1.3)

The type definitions that have been introduced in v7 appear to be too strict and don't allow users to leave out properties that have default values.

As an example, the following code currently throws an error when using it with TypeScript:
```ts
import FakeTimers from "@sinonjs/fake-timers";

FakeTimers.install({
  now: 1
});

FakeTimers.install();
```

This is because the type definitions do not have question marks for optional properties or arguments. Function parameters and object properties can be marked as optional using a question mark, e.g.:
```ts
function example(requiredParam1: { requiredProp1: number; optionalProp1?: number }, optionalParam1?: number) {
  // ...
}
```

**How to reproduce**

You can check out the following codesandbox: https://codesandbox.io/s/headless-lake-jq53p?file=/src/index.ts

geigerzaehler:
#### Purpose (TL;DR)

We make the configuration argument for `install` and all configuration properties optional in the TypeScript definition. This fixes #352.

#### Background

As described in #352, the type definitions generated from the JSDoc comments required the config object with all parameters when calling `install`. All of these parameters are optional however.

#### Solution

It was necessary to create a JSDoc `@typedef` section because TypeScript is not able to generate the intended types from

    @param [config.now] {number|Date}

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -1212,12 +1212,18 @@ function withGlobal(_global) {
     }
 
     /**
-     * @param config {Object} optional config
-     * @param config.now {number|Date}  a number (in milliseconds) or a Date object (default epoch)
-     * @param config.toFake {string[]} names of the methods that should be faked.
-     * @param config.loopLimit {number} the maximum number of timers that will be run when calling runAll()
-     * @param config.shouldAdvanceTime {Boolean} tells FakeTimers to increment mocked time automatically (default false)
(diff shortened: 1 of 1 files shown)


fatso83 (approved):

mroderick:
This has been published to the npm registry as `@sinonjs/fake-timers@7.0.2`

PR #357, opened 2021-01-19
Add docs about typings

vovkasm:
New incomplete typings distributed with version 7 will be selected by typescript. That breaks projects. Add recipe how to guide typescript to select typings from DefinitelyTyped.

* See https://github.com/sinonjs/fake-timers/issues/352#issuecomment-762318145


fatso83 (approved):
🥳

Thanks! This is great.

PR #362, opened 2021-02-05
Add timer getters for returned object via setTimeout, etc.

XadillaX:

Sometimes we want to get inner metadata (e.g. `delay`) of a timer. The original way is to use an undocumented map `clock.timers[timer.id]`.

This PR is to make us can use `timer.delay` directly.

#### Solution

Add some `getter`s in returned object bind to the original timer metadata.


fatso83:
@XadillaX I am trying to keep my hands off this project, so hoping @lolex-core can take a look at this and publish a new version.

fatso83:
@XadillaX I am trying to keep my hands off this project, so hoping @sinonjs/lolex-core can take a look at this and possibly publish a new minor version.

benjamingr:
Fwiw I saw this but I'm not sure it's a good idea to expose this functionality 😅

Sorry for the lack of feedback and dropping the ball... I'm also with a newborn, so I don't have a ton of free time.

XadillaX:
> Fwiw I saw this but I'm not sure it's a good idea to expose this functionality 
> 
> Sorry for the lack of feedback and dropping the ball... I'm also with a newborn, so I don't have a ton of free time.

Sometimes I want to check the delay value of a timer that some function to be tested just set.

benjamingr:
> Sometimes I want to check the delay value of a timer that some function to be tested just set.

Isn't checking the effect (that is, `.tick`ing that duration and then seeing the effect happened) sufficient?

test/fake-timers-test.js
```diff
@@ -277,6 +277,13 @@ describe("FakeTimers", function () {
 
             if (typeof result === "object") {
                 assert.isNumber(result.id);
+
+                var timer = this.clock.timers[result.id];
+                for (var key in timer) {
+                    if (timer.hasOwnProperty(key)) {
+                        assert.equals(result[key], timer[key]);
+                    }
+                }
```
fatso83, line 286: I don't see what these checks has to do with what the test says it is doing?

src/fake-timers-src.js
```diff
@@ -316,6 +315,25 @@ function withGlobal(_global) {
                     return setTimeout(timer.func, timer.delay);
                 },
             };
+
+            var keys = Object.keys(timer);
+            for (var i = 0; i < keys.length; i++) {
+                var key = keys[i];
+                if (timer.hasOwnProperty(key)) {
```
fatso83, line 322: this check seems superflous:
> The Object.keys() method returns an array of a given object's own enumerable property names, iterated in the same order that a normal loop would.

You already know this is owned by the timer.

src/fake-timers-src.js
```diff
@@ -316,6 +315,25 @@ function withGlobal(_global) {
...
+                        Object.defineProperty(res, key, {
+                            get: function () {
+                                return timer[key];
+                            },
+                            configurable: false,
+                            enumerable: true,
+                        });
+                    })(key);
+                    /* eslint-enable */
+                }
```
fatso83, line 334: Seems a bit overkill to expose every field if all you needed was the `delay`, n'est-ce pas?

fatso83 (commented):
Hi, I see this was lying dormant, so I had a look. I kind of second Benji's take that I fail to see a good reason to do this. Exposing the internals seems like it should lead to bad testing practices, but I am willing to reconsider if I see some good use cases.

Some made up examples are fine.

XadillaX:
When I create a timer, I can assert its delay without ticking it.

e.g.

```js
let timer;
if (a) {
  timer = setTimeout(() => {}, 1000);
} else {
  timer = setTimeout(() => {}, 2000);
}
```

fatso83:
Hi, @XadillaX . Thanks for getting back to us!

To me, this seems like checking the implementation, not the effects of said implementation.  That is usually not regarded good testing practice. You want to know if something happens after X ms through its external interface, but not how (it could be two or ten timeouts in between, but that is internal details and none of your business).

So I kind of second this:

> Isn't checking the effect (that is, .ticking that duration and then seeing the effect happened) sufficient?

XadillaX:
> Hi, @XadillaX . Thanks for getting back to us!
> 
> To me, this seems like checking the implementation, not the effects of said implementation. That is usually not regarded good testing practice. You want to know if something happens after X ms through its external interface, but not how (it could be two or ten timeouts in between, but that is internal details and none of your business).
> 
> So I kind of second this:
> 
> > Isn't checking the effect (that is, .ticking that duration and then seeing the effect happened) sufficient?

The example may not be very clear.

Let's set a certain timer (with same callback), but in different delays.

```js
class Checker {
  constructor() {
    this.timer = null;
    this.checked = false;
  }

  check() {
    const time = Date.now();
    this.timer = setTimeout(doCheck.bind(this), 1000 - (time % 1000));
  }

  doCheck() {
    this.checked = true;
  }
}

const checker = new Checker();
checker.check();
const now = Date.now();
assert((checker.timer.delay + now) % 1000 === 0);

// do other tests to test checker.checked
```

PR #366, opened 2021-03-19
Change the protocol for all URLs to https://

remcohaszing:

This is more secure. Also the inconsistency when running `yarn outdated`  bugged me.

#### Solution  - optional

Replace all `http://` URLs with `https://`. I verified all changed URLs work with https.


fatso83 (approved):
Thanks! Not sure what yarn outdated printed, but good whatever it was is fixed now :)

PR #369, opened 2021-04-08
Fix refresh arguments forwarding

### LINKED ISSUE #368: fake-timers 7 doesn't seem to forward setTimeout args
```js
setTimeout((foo) => {
  console.log(foo) // should log "hello"
}, 1, 'hello')
```

Refs: https://github.com/nodejs/undici/pull/675#issuecomment-812918698

dnlup:
#### Purpose (TL;DR)

Refresh wasn't forwarding the arguments passed to setTimeout.
 
Fixes #368

### DIFF
--- a/test/fake-timers-test.js
+++ b/test/fake-timers-test.js
@@ -4655,6 +4655,23 @@ describe("issue #315 - praseInt if delay is not a number", function () {
     });
 });
 
+describe("#368 - timeout.refresh setTimeout arguments", function () {
+  it("should forward  argumnts passed to setTimeout", function () {
+    var clock = FakeTimers.install();
+    var stub = sinon.stub();
+
+    if (typeof setTimeout(NOOP, 0) === "object") {
+    var t = setTimeout(stub, 1000, "test");
+      clock.tick(1000);
+      t.refresh();
+      clock.tick(1000);
+      assert.calledTwice(stub);
+      assert.alwaysCalledWith(stub, "test");
+    }
+    clock.uninstall();
+  });
+});
+
 describe("#187 - Support timeout.refresh in node environments", function () {
     it("calls the stub again after refreshing the timeout", function () {
         var clock = FakeTimers.install();
--- a/test/fake-timers-test.js
+++ b/test/fake-timers-test.js
@@ -4656,20 +4656,23 @@ describe("issue #315 - praseInt if delay is not a number", function () {
 });
 
 describe("#368 - timeout.refresh setTimeout arguments", function () {
-  it("should forward  arguments passed to setTimeout", function () {
-    var clock = FakeTimers.install();
-    var stub = sinon.stub();
+    before(function () {
+        if (!addTimerReturnsObject) {
+            this.skip();
+        }
+    });
+    it("should forward  arguments passed to setTimeout", function () {
+        var clock = FakeTimers.install();
+        var stub = sinon.stub();
 
-    if (typeof setTimeout(NOOP, 0) === "object") {
-    var t = setTimeout(stub, 1000, "test");
-      clock.tick(1000);
-      t.refresh();
-      clock.tick(1000);
-      assert.calledTwice(stub);
-      assert.alwaysCalledWith(stub, "test");
-    }
-    clock.uninstall();
-  });
+        var t = setTimeout(stub, 1000, "test");
+        clock.tick(1000);
+        t.refresh();
+        clock.tick(1000);
+        assert.calledTwice(stub);
+        assert.alwaysCalledWith(stub, "test");
+        clock.uninstall();
+    });
 });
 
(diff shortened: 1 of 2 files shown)


test/fake-timers-test.js
```diff
@@ -4655,6 +4655,23 @@ describe("issue #315 - praseInt if delay is not a number", function () {
     });
 });
 
+describe("#368 - timeout.refresh setTimeout arguments", function () {
+  it("should forward  argumnts passed to setTimeout", function () {
```
fatso83, line 4659:
```suggestion
  it("should forward  arguments passed to setTimeout", function () {
```

test/fake-timers-test.js
```diff
@@ -4655,6 +4655,23 @@ describe("issue #315 - praseInt if delay is not a number", function () {
...
+    var clock = FakeTimers.install();
+    var stub = sinon.stub();
+
+    if (typeof setTimeout(NOOP, 0) === "object") {
```
fatso83, line 4663: Why is this needed? Is this some kind of check for whether the setTimeout object was installed or not? I think this should be removed, as the test would otherwise crash (rightfully so). If the timer is not installed, then we have other issues, TBH :)
```suggestion
```

Suggest removing this, but I might be not understanding something obvious here (,but then add a comment at least :) )

test/fake-timers-test.js
```diff
@@ -4655,6 +4655,23 @@ describe("issue #315 - praseInt if delay is not a number", function () {
...
+    var t = setTimeout(stub, 1000, "test");
+      clock.tick(1000);
+      t.refresh();
+      clock.tick(1000);
+      assert.calledTwice(stub);
+      assert.alwaysCalledWith(stub, "test");
+    }
```
fatso83, line 4670:
```suggestion
```

And run `npm run prettier:write` afterwards :)

fatso83 (changes requested):
Thanks for the fix! Just did not understand the guard in the test.

test/fake-timers-test.js
dnlup: I agree with you. I just saw this in other tests and thought it was needed. Without it, the test crashes. But only on the cli, if I run it with `--inspect-brk` and open the dev tools it doesn't. I don't understand this either.

benjamingr: > Why is this needed? Is this some kind of check for whether the setTimeout object was installed or not?

This is a check if you have Node.js timers (objects) or web timers (numbers)

benjamingr: Because `.refresh` does not exist on web timers

benjamingr (commented):
Changes look good, thanks :) 

(I'll approve after the comments are addressed since Carl's comments are good)

test/fake-timers-test.js
mantoni: It would be more obvious if the test was conditionally skipped, like be do here:

https://github.com/sinonjs/fake-timers/blob/54852f6d270295f164e7c1d8c217955523e22cfb/test/fake-timers-test.js#L450-L454

fatso83: Yeah, we usually do feature detection and skip tests (or a whole suite) by using Mocha's `Context#skip()` feature. That makes each test more succinct and easier to read.

dnlup: > Because `.refresh` does not exist on web timers

Thanks for the clarification

dnlup: > It would be more obvious if the test was conditionally skipped, like we do here:
> 
> https://github.com/sinonjs/fake-timers/blob/54852f6d270295f164e7c1d8c217955523e22cfb/test/fake-timers-test.js#L450-L454

I'll change it. Thank you for the suggestion.

dnlup:
I think we are there if I haven't missed anything.

test/fake-timers-test.js
```diff
@@ -4655,6 +4655,26 @@ describe("issue #315 - praseInt if delay is not a number", function () {
...
+            this.skip();
+        }
+    });
+    it("should forward  arguments passed to setTimeout", function () {
+        var clock = FakeTimers.install();
+        var stub = sinon.stub();
+
+        var t = setTimeout(stub, 1000, "test");
+        clock.tick(1000);
+        t.refresh();
```
benjamingr, line 4670: Does this pass in browsers?

dnlup: I ran:

```bash
$ npm test
$ npm run test-headless
```

Should I run some other script?

benjamingr: That's probably a _separate_ bug (refresh existing when run in browsers), heh.

benjamingr (approved):

mroderick:
This has been published to the npm registry as part of `@sinonjs/fake-timers@7.0.5`

PR #370, opened 2021-04-09
Fix up some typescript types

43081j:
#### Purpose (TL;DR) - mandatory

Just trying to clean up some mistakes in the JSDoc types.

These are the only problems im aware of. I will do another PR at some point that improves the types if i can (e.g. getting rid of the any).

cc @fatso83

PR #371, opened 2021-04-09
Add further stronger typing and depend on globalThis

43081j:

Continued from #370.

I've tried to do the following:

* Depend on `globalThis`, such that things like `setTimeout` are strongly typed as being identical to `typeof globalThis.setTimeout` - this should theoretically mean you get the right types depending on the environment you're in (node, browser)
* Define node-specific functions ourselves to avoid pulling node types in
* Define what `timers`
* Add jsdoc to `withGlobal` but _not_ a return type (so typescript will infer it)

If someone can double check that depending on the global types makes sense and will work fine, that'd be much appreciated. It seems to for me

cc @fatso83

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -6,7 +6,7 @@ var globalObject = require("@sinonjs/commons").global;
  * @typedef {object} Clock
  * @property {number} now
  * @property {any} timeouts
- * @property {typeof Date} Date
+ * @property {typeof globalThis.Date} Date
  * @property {number} loopLimit
  * @property {(func: Function, timeout: number) => number} requestIdleCallback
  * @property {(timerId: number) => void} cancelIdleCallback
@@ -669,7 +669,7 @@ function withGlobal(_global) {
      * @property {clearTimeout} clearTimeout
      * @property {setInterval} setInterval
      * @property {clearInterval} clearInterval
-     * @property {typeof Date} Date
+     * @property {typeof globalThis.Date} Date
      * @property {((fn: (...args: any[]) => void, ...args: any[]) => NodeTimer)=} setImmediate
      * @property {((id: NodeTimer) => void)=} clearImmediate
      * @property {((time?: [number, number]) => [number, number])=} hrtime
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -6,21 +6,21 @@ var globalObject = require("@sinonjs/commons").global;
  * @typedef {object} Clock
  * @property {number} now
  * @property {any} timeouts
- * @property {any} Date
+ * @property {typeof Date} Date
  * @property {number} loopLimit
- * @property {(func: Function, timeout: number) => any} requestIdleCallback
- * @property {(timerId: number) => any} cancelIdleCallback
- * @property {(func: Function, timerId: number) => any} setTimeout
- * @property {(timerId: number) => any} clearTimeout
- * @property {(func: Function) => any} nextTick
- * @property {(func: Function) => any} queueMicrotask
- * @property {(func: Function, timeout: number) => any} setInterval
- * @property {(timerId: number) => any} clearInterval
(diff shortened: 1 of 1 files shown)


fatso83:
Thanks for this! I see there is a big discussion leading up to this on DT and I think we are well aware that the types we ship are not as up to snuff as the ones on DT, although we hope that might change in time :)

> If someone can double check that depending on the global types makes sense and will work fine

TBH, I am not sure how to verify that, given that I don't have any typescript consumers at the moment, but a manual inspection of this seems to imply these changes are beneficial ...

43081j:
fwiw i ran the build and manually checked the produced types by eye, which is how i got to these changes.

tomorrow if i get chance, i will run the DT tests against them and that should be enough to verify if they are fine IMO.

its also a major version bump over on DT so if we accidentally weaken some types, thats ok, ill fix them up in separate PRs.

src/fake-timers-src.js
```diff
@@ -33,13 +33,37 @@ var globalObject = require("@sinonjs/commons").global;
...
-// eslint-disable-next-line complexity
+/**
+ * Configuration object for the `install` method.
+ *
+ * @typedef {object} Config
+ * @property {number|Date} now a number (in milliseconds) or a Date object (default epoch)
+ * @property {string[]} toFake names of the methods that should be faked.
+ * @property {number} loopLimit the maximum number of timers that will be run when calling runAll()
+ * @property {Boolean} shouldAdvanceTime tells FakeTimers to increment mocked time automatically (default false)
+ * @property {Number} advanceTimeDelta increment mocked time every <<advanceTimeDelta>> ms (default: 20ms)
```
stof, line 50: shouldn't this be `number` rather than `Number` (i.e. asking for a primitive number rather than a boxed Number object) ? Same for `boolean` on the previous line.

The typescription documentation about writing types says that you almost never want to use the boxed types in type declarations.

fatso83: Pretty sure we want to use primitives rather than boxed types everywhere. That also aligns with how I read [the JSDoc docs](https://jsdoc.app/tags-type.html)

43081j: Yep it's just a typo. Have spent too much time writing the constructors recently in other projects and had it stuck in my head 😂

43081j:
@stof @fatso83 have pushed the fixes to those constructor types, mind taking another look?

43081j:
one thing to note here that i need double checking:

```ts
let num: number = clock.setTimeout(fn, 1000);

// in a browser, this is a `number`
// in node, it is a `NodeTimer`, an object
// this is expected and fine.

// However, the following won't work:
num = clock.setImmediate(fn);

// This is a `NodeTimer`, always, as setImmediate doesn't exist in browsers.
// It can't be assigned to a `number`
```

This shouldn't ever happen but we can't really type it easily through JSDoc.

In reality, the `Clock` type would _never_ have a `setImmediate` in browsers but would always have one in node. Making the above code impossible in browsers.

Because its not typed that way, there's nothing stopping a browser consumer trying to use `setImmediate`, but it wouldn't really exist

fatso83 (approved):

fatso83:
> the Clock type would never have a setImmediate in browsers 

Actually, that is not _entirely_ true. I have used `setImmediate` lots of times: it was in [IE10, IE11 and Edge 12-18](https://caniuse.com/?search=setImmediate) and polyfills were available for all the rest. Now that the world, including MS has essentially gone mono-browser (via Chrome) it is _practically_ true for evergreen browsers 😄 

TypeScript does support [Conditional Typing](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html), so it's possible to rectify this with a 10/10 solution at some time, but I think your solution is good enough to merge now.

43081j:
@fatso83 could you possibly publish a new version with this change in? So we can unblock the last change in the DT repo

PR #373, opened 2021-04-15
chore: add type tests

SimenB:

https://github.com/sinonjs/fake-timers/issues/356#issuecomment-768227159

I just added a dummy test, no attempt at coverage. But it's a good opportunity for new contributors if they want, and future changes to the types should include tests

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -20,3 +20,9 @@ overrides:
     rules:
       max-nested-callbacks: off
       mocha/no-exclusive-tests: error
+  - files: "**/*.ts"
+    parserOptions:
+      ecmaVersion: 2020
+      sourceType: 'module'
+    rules:
+      no-unused-expressions: off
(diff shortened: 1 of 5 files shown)


.eslintrc.yaml
```diff
@@ -20,3 +20,9 @@ overrides:
     rules:
       max-nested-callbacks: off
       mocha/no-exclusive-tests: error
+  - files: "**/*.ts"
+    parserOptions:
+      ecmaVersion: 2020
+      sourceType: 'module'
```
SimenB, line 26: removes the syntax error for `import`, but still syntax error for `type`. Might be worth it to add `@typescript-eslint/parser` or something, but felt very out of scope

mroderick:
This looks like a good start to me. @SimenB can you think of any reason not to merge this?

mroderick (approved):

SimenB:
Only reason would be if we wanted more tests before landing, but I think it's valuable as is 👍

benjamingr (approved):

PR #375, opened 2021-05-01
Retry - Add stack trace to code recursively scheduling timers #325 

itayperry:
Re-adding the reverted merge - Add stack trace to code recursively scheduling timers #325 

I've cherry-picked commits made by @alistairjcbrown to re-file the original PR that was merged and then reverted due to some compatibility issues.

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -252,30 +252,30 @@ function withGlobal(_global) {
             `Aborting after running ${clock.loopLimit} timers, assuming an infinite loop!`
         );
 
-        var clockMethodPattern;
-
-        // pattern never matched in Firefox and Node
+        // pattern never matched in Node
         var computedTargetPattern = /target\.*[<|(|[].*?[>|\]|)]\s*/;
+        var clockMethodPattern = new RegExp(
+            String(Object.keys(clock).join("|"))
+        );
 
         if (addTimerReturnsObject) {
             // node.js environment
             clockMethodPattern = new RegExp(
                 `\\s+at (Object\\.)?(?:${Object.keys(clock).join("|")})\\s+`
             );
-        } else {
-            clockMethodPattern = new RegExp(
-                String(Object.keys(clock).join("|"))
-            );
         }
 
         var matchedLineIndex = -1;
         job.error.stack.split("\n").some(function (line, i) {
-            // If we've matched a computed target line (e.g. setTimeout) then we
-            // don't need to look any further. Return true to stop iterating.
-            var matchedComputedTarget = line.match(computedTargetPattern);
-            if (matchedComputedTarget) {
-                matchedLineIndex = i;
-                return true;
+            if (!addTimerReturnsObject) {
+                // when not in node
+                // If we've matched a computed target line (e.g. setTimeout) then we
+                // don't need to look any further. Return true to stop iterating.
+                var matchedComputedTarget = line.match(computedTargetPattern);
+                if (matchedComputedTarget) {
+                    matchedLineIndex = i;
+                    return true;
+                }
             }
 
             // If we've matched a clock method line, then there may still be
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -267,15 +267,13 @@ function withGlobal(_global) {
 
         var matchedLineIndex = -1;
         job.error.stack.split("\n").some(function (line, i) {
-            if (!addTimerReturnsObject) {
-                // when not in node
-                // If we've matched a computed target line (e.g. setTimeout) then we
-                // don't need to look any further. Return true to stop iterating.
-                var matchedComputedTarget = line.match(computedTargetPattern);
-                if (matchedComputedTarget) {
-                    matchedLineIndex = i;
-                    return true;
-                }
+            // If we've matched a computed target line (e.g. setTimeout) then we
+            // don't need to look any further. Return true to stop iterating.
+            var matchedComputedTarget = line.match(computedTargetPattern);
+            /* istanbul ignore if */
+            if (matchedComputedTarget) {
+                matchedLineIndex = i;
+                return true;
             }
 
             // If we've matched a clock method line, then there may still be
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -152,6 +152,18 @@ function withGlobal(_global) {
         return isFinite(num);
     }
 
+    var isNearInfiniteLimit = false;
+
+    function checkIsNearInfiniteLimit(clock, i) {
+        if (clock.loopLimit && i === clock.loopLimit - 1) {
+            isNearInfiniteLimit = true;
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


fatso83:
So [this fails in IE 11, Safari 9, Firefox](https://github.com/sinonjs/fake-timers/files/6413732/out.txt). This ties in with https://github.com/sinonjs/sinon/issues/2359 which is about which browsers we officially support and that our docs is out-of-sync. We have dropped IE 11 and Safari 9 in favor of evergreen browsers, so we could ignore them here as well, but it fails in Firefox, which we do support, and we need to handle that somehow.

fatso83:
IE11 and Safari has been dropped. Any chance of seeing if you can get this running in Firefox as well?

itayperry:
Yeah sure @fatso83, I don't know how to do it, but I would like to try! :) 
I'm on it.

itayperry:
Thank you @fatso83 for helping out and explaining everything!  

> just copy-pasting the entire script

Do you mean this script:
`"test-cloud": "mochify --wd --no-detect-globals --timeout=10000"` that appears in `package.json`?

I'm just guessing because this script shows `Unexpected HTTP status: 401 Unauthorized` when running it, and because it's the only one that I could see using the `.min-wd` config file.

fatso83:
I was simply thinking that if you needed to test certain things in Firefox that failed in the Sauce Labs run you could just copy-paste the entire fake-timers.js script and just manually execute whatever failed in the test runs.

The `test-cloud` script returns 401 since you do not have credentials setup. I contacted you on Twitter to send you the credentials. You need `$SAUCE_ACCESS_KEY ` and `$SAUCE_USERNAME` in your environment. If you pass me an email (on my profile) you can get it right away.

itayperry:
I didn't understand the script way at first, so I decided to try the other way with the local wd.
I'm just a little slow on the take right now :)
If I wanna use the script in the console should I build it first?

And thank you for the email!! 🌼 and for all the help.

fatso83:
> If I wanna use the script in the console should I build it first?

As [the docs say](https://github.com/sinonjs/fake-timers#installation), you can either build it or use Skypack. Since the latter assumes it's already on NPM, you need to build it for the browser: 
```
npx browserify src/fake-timers-src.js -o lolex.js
```

If you want all tests in the browser:
```
npx browserify test/fake-timers-test.js -o tests.js
```

itayperry:
Hi @fatso83, thanks for explaining - I tried both:

`npx browserify src/fake-timers-src.js -o lolex.js` worked (I think), the second one threw an error.

I'm gonna try to debug the script, do you know how do I access `createClock()` and basically all the other methods? I've looked it up on the 'window' object after running the script, but I couldn't find it. I don't know the naming in which it was saved.

itayperry:
Hi @fatso83 (and @benjamingr ) :) 
All the tests now pass, I just pushed the fix, when you're available please update me if everything is really ok. 

Tons of RegExp lol my head is spiraling!

itayperry:
BTW, I'm trying to run `npm run prettier:write` and I don't know why it fails.. 
prettier-error

fatso83:
@itayperry I see why it fails! It's because we are kind of a unix shop, and I see the `"C:\Projects\fake-timers"` string indicating you are not in that camp 😄 So NPM will run its script in the default shell on your machine. The default shell on your machine is `CMD.exe`, which does not know the `{js,css,ms}` syntax at all. It thinks it is a verbatim string to match.

So ... either accept that you need to do a manual `prettier --write */*.js` (or whatever syntax works on Windows) OR check out [WSL2](https://docs.microsoft.com/en-us/windows/wsl/install-win10), which works great (I am developing on Windows these days using mainly WSL2 to get all my Unix fixes covered). If you just check out the project on [WSL2 ](https://docs.microsoft.com/en-us/windows/wsl/install-win10)it will "just work".

fatso83:
P.S. I have not checked out your code yet, but just a general thing: instead of trying to create One Regex To Rule Them All ™️ , you can try to split it up into different "branches" and test something depending on what matches some substring. Just to get the complexity down.

itayperry:
> The default shell on your machine is CMD.exe, which does not know the {js,css,ms} syntax at all. It thinks it is a verbatim string to match.

ohhh now I get it! thank you :) I've used the WSL2 once, I'll try and use it again after I finish this PR.

>  test something depending on what matches some substring. Just to get the complexity down.

I didn't think about it actually... RegExp can be quite intimidating sometimes for me. 
For now, what I did was to add an `if` to check whether the env is Node.js

itayperry:
Update: now all the tests pass, which is wonderful :)

eslint yelled at me lol - it didn't work at all and I think it was updated or something so I tried `npm install` and now all hell broke loose.. I'm getting many warnings - even about code that's not in the PR.

I fixed the errors but there are still approximately 100 warnings.

itayperry:
Perhaps I should use something like this: `/* istanbul ignore if */` ?

fatso83:
Does this mean it actually runs in Firefox?

itayperry:
@fatso83 Yes, it runs in Firefox, and from what I remember also Chrome :)

fatso83:
To me this sound like it's ready to merge :)

SimenB:
@fatso83 can you release this? 🙂 (I guess major after #385 etc)

fatso83:
Hi, @SimenB . Welcome back :smiley_cat: 
I actually thought that both you and Benjamin both had NPM release rights. Could arrange for that, rather than having you wait for me.

Not totally sure if there is something else breaking worth thinking about ... not quite sure why #385 would be breaking though. We have [targetted Evergreen browsers for some time](https://github.com/sinonjs/sinon/blob/master/COMPATIBILITY.md), so if ES2015+ features have snuck in, that's to be expected. My memory is old and faded, though :shrug:

SimenB:
Ah, even better 😀

fatso83:
@SimenB I released it as v8.0.0 now, since I was a bit unsure about how the removal of TS bindings would affect current installs. Someone might see some things break, so better safe than sorry.

SimenB:
Thanks!

fatso83:
Bare hyggelig :norway:

SimenB:
Funka nesten 😀 

There's a bug here, I'm seeing `Cannot read property 'stack' of undefined` - will dig into this and send a PR

fatso83:
amagad, should we mark v8 as deprecated and move the latest tag to v7.x?

SimenB:
Yeah, probably

fatso83:
Ah, you do have rights already. 
```
$ npm show @sinonjs/fake-timers maintainers
 
```json
[
  'simenb <sbekkhus91@gmail.com>',
  'mrgnrdrck <morgan@roderick.dk>',
  'fatso83 <carlerik@gmail.com>',
  'mantoni <mail@maxantoni.de>',
  'cjohansen <christian@cjohansen.no>',
  'benjamin.gruenbaum <benjamingr@gmail.com>'
]
```

PR #376, opened 2021-05-03
Drop IE11 and Safari 9 from Sauce Labs test runners

fatso83:

- Drop IE11 and Safari 9 per discussion in #375 and sinonjs/sinon#2359.
- Also copies in browsers from the mother Sinon repo to include Edge and Safari latest

See discussion in #375 and sinonjs/sinon#2359

### DIFF
--- a/.min-wd
+++ b/.min-wd
@@ -2,5 +2,9 @@
   "sauceLabs": true,
   "browsers": [{
     "name": "firefox"
+  }, {
+    "name": "MicrosoftEdge"
+  }, {
+    "name": "safari"
   }]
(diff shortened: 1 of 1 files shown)


benjamingr:
Should we do a version that modernizes the codebase? (I'm happy to run the code through transforms that modernize it if we don't need to support IE11)

fatso83:
I would not mind that. The only problem I have with large touch-everything stuff like running Prettier, Eslint and transforms on the code is that it makes `git blame` a lot less useful, but I can live with that.

PR #381, opened 2021-05-27
Fix JSDoc for createClock

henhal:
This improves the generated .d.ts so that createClock may be called without arguments.

Improve .d.ts types for `createClock()` so that it's possible to call it from TypeScript without arguments.

The currently generated .d.ts uses JSDoc, which incorrectly specified the two arguments for `createClock()` as mandatory.

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -731,8 +731,8 @@ function withGlobal(_global) {
     var originalSetTimeout = _global.setImmediate || _global.setTimeout;
 
     /**
-     * @param {Date|number} start the system time - non-integer values are floored
-     * @param {number} loopLimit maximum number of timers that will be run when calling runAll()
+     * @param {Date|number} [start] the system time - non-integer values are floored
+     * @param {number} [loopLimit] maximum number of timers that will be run when calling runAll()
      * @returns {Clock}
      */
     function createClock(start, loopLimit) {
@@ -1309,7 +1309,7 @@ function withGlobal(_global) {
     /* eslint-disable complexity */
 
     /**
-     * @param {Config=} config Optional config
+     * @param {Config=} [config] Optional config
      * @returns {Clock}
      */
     function install(config) {


henhal:
Authors: IMHO, generating .d.ts from JSDoc should be considered a very temporary approach. Is there a reason why this was introduced and put in favour of DefnitelyTyped? Now when importing the library it's _harder_ to use than before, since the generated typings are not at all mature. The _very first_ call I made was to `createClock()`, and I was confused by it requiring two parameters...

I'd love to help, but I noticed that there would be quite a lot of work to fix all the function types etc. To get this working by only improving the JSDoc is going to be harder than any other solution I think. Actually I don't know in which shape DefinitelyTyped is for this repo, but I'm not sure it's a step in the right direction to go from a community managed DefinitelyTyped repo to an auto-generated one that suffers from poor JSDoc conversion.

In my book, of course the best way would be to rewrite Sinon in TypeScript. Are contributions for that of interest?

The second best is probably to add the d.ts files to source control and manually improve them within this repository.

The third best is to rely on DefinitelyTyped.

The worst, IMHO, is the current approach...

henhal:
Perhaps it can also be noted that the suggest approach for still using DefinitelyTyped does not work perfectly either. If modifying `paths` in `tsconfig.json`, my code compiles OK, and clicking on symbols from the library takes me to the DT repo. However, the IDE (Webstorm 2020) still links to the d.ts file in the source repo and refuses to understand that the code does compile. 

This may be an IDE issue of course, but it still highlights that this library became much harder to use than before. I ended up downgrading to ^6.0.1. :(

fatso83:
Thanks for the PR! As to your comments, I'll leave that for now (as I'm in a hurry).

fatso83:
>  Is there a reason why this was introduced and put in favour of DefnitelyTyped? 

Of course. If you are interested in the background you can find most of the reasoning in the discussion here: https://github.com/sinonjs/fake-timers/issues/220#issuecomment-430596244

> go from a community managed DefinitelyTyped repo to an auto-generated one that suffers from poor JSDoc conversion.

We go from an external repo that is usually out of sync (I have contributed to the DT types several times when stuff I need at work is missing from DT) to one managed _by the community_. The main point for us is documenting the types used _using JSDoc_. The generated TypeScript definitions are a _bonus_.

> In my book, of course the best way would be to rewrite Sinon in TypeScript. Are contributions for that of interest?

Not at all, but not from bad will 😄  You need maintainers willing to keep it up to date in Typescript - of which there are none. It was hard enough getting contributors _before_ TS was a thing. We have no interest in Babel, bundlers and source maps getting in our way when debugging - I think we all have enough battle scars from the last decade of fighting various generations of such machinery 😸

fatso83:
This is related to ☝️ : #386

PR #392, opened 2021-06-18
Const for var

fatso83:

Use one of the simpler ES2015+ features.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "license": "BSD-3-Clause",
   "scripts": {
     "lint": "eslint .",
-    "test-node": "mocha test/ integration-test/ -R dot --check-leaks",
+    "test-node": "mocha --timeout 200 test/ integration-test/ -R dot --check-leaks",
     "test-headless": "mochify --no-detect-globals --timeout=10000",
     "test-check-coverage": "npm run test-coverage && nyc check-coverage",
     "test-cloud": "mochify --wd --no-detect-globals --timeout=10000",
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -1,6 +1,6 @@
 "use strict";
 
-var globalObject = require("@sinonjs/commons").global;
+const globalObject = require("@sinonjs/commons").global;
 
 /**
  * @typedef {object} IdleDeadline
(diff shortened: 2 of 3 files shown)


benjamingr:
We can probably pretty easily modernise the codebase - I thought it was ES5 "for a reason"  but there are plenty of transforms/tools we can use to transform it to ES6 (like lebab iirc).

fatso83:
There was a reason and that reason went away with Sinon 10 that dropped all IE11 compatibility. We are now in in the subset of evergreen+Safari territory 😃 Feel free to do some auto-magic on the rest, but it did require a bit of manual fiddling and test reruns to get the `const`/`let` right.

package.json
```diff
@@ -15,7 +15,7 @@
   "license": "BSD-3-Clause",
   "scripts": {
     "lint": "eslint .",
-    "test-node": "mocha test/ integration-test/ -R dot --check-leaks",
+    "test-node": "mocha --timeout 200 test/ integration-test/ -R dot --check-leaks",
```
mroderick, line 18: That's a pretty big smell from replacing `var` with `const` and `let`.

Is this really necessary?

fatso83: Not sure I agree: that was just an improvement of our general setup that I stumbled upon when doing this. Without it, if you should remove the `done` callback in any of our tests it would just hang for 2 whole seconds, which was a PITA when doing this, as so many tests were breaking. None of our tests should use anywhere near 100 ms, but I was just going conservative when setting it to 200 ms. 

Before:
```
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/carlerik/dev/fake-timers/test/fake-timers-test.js)
```

After:
```
     Error: Timeout of 200ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/carlerik/dev/fake-timers/test/fake-timers-test.js)
```

We could of course stuff it in a .mocharc or something, but we do not really have _that_ many options to hide. What do you suggest?

mroderick: When I read it, I thought that we would accept tests running for `200ms` as acceptable. Which seems excessive.

I guess that's what we're actually doing though.

Thank you for the explanation 👍

Issue #380, opened 2021-05-20
Safari test run breaks in Sauce Lab due to missing W3C WebDriver support


fatso83:
Full background: https://github.com/sinonjs/sinon/issues/2372.

Basically, Safari tests breaks with v12 and up. This needs to be fixed in `min-wd` (mantoni/min-webdriver#31).

benjamingr:
Would it make sense to just migrate to WebDriverIO? 

@christian-bromann given the visibility and popularity of this project is this something you can assist with?
(feel free to say "no", no intention to pressure)

Ideally someone looking to contribute more ( @itayperry ) could take this if you can mentor them on the migration.

christian-bromann:
> Would it make sense to just migrate to WebDriverIO?

I don't think so. WebdriverIO has no capabilities to run Mocha, Jasmine and Cucumber in a browser for unit test purposes. However given that many libraries still would like to do such things I will look into supporting this. Short term I believe fixing the protocol bug in `min-webdriver` makes the most sense IMO.

benjamingr:
@christian-bromann thanks for the quick response it's appreciated! 🙏

The logic for running mocha tests in browsers isn't in min-webdriver it's in mochify. That could (at least optionally?) just use WebdriverIO right?

christian-bromann:
Yes it could. Then easiest would be to implement a browserify plugin compatible interface into `webdriverio` similar as `min-wd` does it.

benjamingr:
Great, @itayperry is this something you'd be interested in contributing? It could be a good chance to contribute to another project too (WebdriverIO)

itayperry:
Yes, I'd be happy - I'm still working on #375 and I also wanna contribute to #374.
I've been slow lately, so if no one takes this one, I definitely will :) I don't wanna stall you.

itayperry:
> ... to implement a browserify plugin compatible interface into `webdriverio` similar as `min-wd` does it

I'd be happy to do it - how / where can I get started? I never did this before but it sounds interesting! :)

mantoni:
I'm actually working on a Mochify rewrite (https://github.com/mantoni/mochify.js has a `rewrite` branch), and the current state seems to work fine. I just don't have enough free time to get it across the finish line.

mantoni:
@itayperry You could start by checking out this branch and see if you can run Safari tests for fake-timers with it: https://github.com/mantoni/mochify.js/pull/241

itayperry:
[...]

PR #395, opened 2021-09-04
New mochify

mantoni:

I'm working on a rewrite of mochify which aims to solve #380, is more modular and works with any bundler.

The current mochify implementation is dated and does not support the latest webdriver protocols.

#### Solution  - optional

The new mochify implementation works with pluggable drivers. The `@mochify/driver-webdriver` uses the official `webdriver` module internally.

~~Interestingly, this successfully runs tests in Safari (although there are failing tests), but I'm having trouble running tests in Firefox and Edge.~~

New Mochify development happens on the `rewrite` branch here: https://github.com/mantoni/mochify.js/tree/rewrite

### DIFF
--- a/mochify.webdriver.js
+++ b/mochify.webdriver.js
@@ -1,0 +1,25 @@
+"use strict";
+
+module.exports = {
+    driver: "webdriver",
+    // eslint-disable-next-line camelcase
+    driver_options: {
+        // Tests must run on a http website for Microsoft Edge to have
+        // sessionStorage.
+        url: "http://maxantoni.de/doctype.html",
+        hostname: "ondemand.saucelabs.com",
+        path: "/wd/hub",
+        port: 80,
+        capabilities: {
+            browserName: process.env.BROWSER_NAME,
+            browserVersion: "latest",
+            "sauce:options": {
+                build: `${process.env.BROWSER_NAME} ${
+                    process.env.GITHUB_RUN_NUMBER || Date.now()
+                }`,
+                username: process.env.SAUCE_USERNAME,
+                accessKey: process.env.SAUCE_ACCESS_KEY,
+            },
+        },
+    },
+};
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -79,10 +79,10 @@ jobs:
                   PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: 1
             - name: Integration
               run: |
-                  export SINON_CHROME_BIN=$(which google-chrome-stable)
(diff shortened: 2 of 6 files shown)


fatso83:
Only thing that is failing is the missing Chromium:
```
Run export SINON_CHROME_BIN=$(which google-chrome-stable)

> @sinonjs/fake-timers@7.1.2 test-headless
> mochify --driver puppeteer "--chrome" "/usr/bin/google-chrome-stable" "--allow-chrome-as-root"

Error: Could not find expected browser (chrome) locally. Run `npm install` to download the correct Chromium revision (901912).
```

mantoni:
Builds now pass, except for Safari which has failing unit tests.

fatso83:
Any idea what the "kl" in the output is? Some debugging statement left in?
```
> @sinonjs/fake-timers@8.0.0 test-headless /Users/carlerik/dev/fake-timers
> mochify --driver puppeteer

kl

  .................................,,,,,,,,,,,,,,,.........................................................................................................
  .............................................,...,...........,...........................,,...................................,..............,......,,,,,
  ,,,,,,...,,,,,,,,,,,,,,,,,..........,..,,........,,..

  304 passing (5s)
  55 pending
```

mantoni:
> Any idea what the "kl" in the output is?

When I run it locally with the current state of mochify linked locally, this doesn't appear.

fatso83:
Hah, probably me that has fiddled around with node_modules when debugging at some earlier stage and forgot 😄

fatso83:
Just wondering about the tests that fail in Safari ... how do I run these Mochify tests locally against a local Safari instance?

mantoni:
You can launch `safaridriver` locally. It ships with Safari.

Setup and config examples are documented here: https://github.com/mantoni/mochify.js/blob/rewrite/driver-webdriver/README.md

mantoni:
I finally manager to get a new Mochify release out and updated this PR.

I have no idea why the selenium job is skipped. Anyone?

Happy Christmas 🎄

mochify.webdriver.js
```diff
@@ -0,0 +1,25 @@
+"use strict";
```
SimenB, line 1: we should ignore this file when collecting code coverage

mantoni:
After a small adjustment, tests now also pass in all browsers, including Safari. Unfortunately Safari does not include the function name in the `err.stack`, so I had to make some assertions less specific.

While testing I also removed Browserify and used ESBuild instead. If this is not wanted, I'm happy to switch it back.

SimenB (approved):

mroderick (approved):
Very nice 👍

PR #398, opened 2021-09-15
chore: add failing test for loop limit error

SimenB:

Failing test from https://github.com/sinonjs/fake-timers/pull/375#issuecomment-919976426.

By not resetting, the counter is not tied to a specific clock but rather then entire `withGlobal`.

I'll push a second commit moving the attribute from living in the `withGlobal` closure to living on the `clock` object, which I think is more correct


SimenB:
Pushed a commit that's _almost_ there - `interval` fails since the timer itself doesn't have an error object - only the last one does. Not sure how to deal with it? A single interval will run forever with `runAll` without us ever registering an `error`

src/fake-timers-src.js
```diff
@@ -1396,18 +1399,18 @@ function withGlobal(_global) {
...
                 numTimers = Object.keys(clock.timers).length;
                 if (numTimers === 0) {
-                    resetIsNearInfiniteLimit();
+                    resetIsNearInfiniteLimit(clock);
                     return clock.now;
                 }
 
-                clock.next();
                 checkIsNearInfiniteLimit(clock, i);
+                clock.next();
```
SimenB, line 1413: moving this after the check fixes all but the `setInterval` tests

SimenB:
@fatso83 I don't know how to deal with `setInterval`. Ideas?

One quick fix is of course to handle `error` not being present in `getInfiniteLoopError` and just return the error created there without trying to get a better trace

SimenB:
Pushed that latest thing - not ideal, but I don't know how to solve it properly

fatso83:
This looks fine, but TBH I am just too tired to grok the details of this right now, so I cannot really say something very useful about the `.error` prop  ...  I was kind of wondering why we don't check for the max number of iterations in that timer to begin with?

benjamingr (approved):

test/fake-timers-test.js
```diff
@@ -5036,6 +5036,7 @@ describe("loop limit stack trace", function () {
         });
     });
 
+    // This doesn't really work since we're unable to add an error to all running intervals
```
fatso83, line 5039: Not totally sure what "this" refers to really. The tests or the `setInteval` function or what.

SimenB: tracking of callbacks triggered by `setInterval`

fatso83:
Is this going somewhere? Can I help? Not totally sure what to do between your last commend and Benji's approval.

PR #401, opened 2021-09-21
fix: handle missing stack when throwing recursion errors

SimenB:
This is for #398 which I think we should land and release as 8.0.1. Guards against the error and falls back to the behavior of v7

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -319,7 +319,7 @@ function withGlobal(_global) {
         );
 
         if (!job.error) {
-            return infiniteLoopError
+            return infiniteLoopError;
         }
 
         // pattern never matched in Node
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -109,6 +109,7 @@ const globalObject = require("@sinonjs/commons").global;
  * @property {number} createdAt
  * @property {boolean} immediate
  * @property {number} id
+ * @property {Error} [error]
  */
 
 /**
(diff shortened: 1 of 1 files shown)


src/fake-timers-src.js
```diff
@@ -109,6 +109,7 @@ const globalObject = require("@sinonjs/commons").global;
  * @property {number} createdAt
  * @property {boolean} immediate
  * @property {number} id
+ * @property {Error} [error]
```
SimenB, line 112: I _think_ this is how you say `optional`?

benjamingr: I think so

src/fake-timers-src.js
```diff
@@ -309,14 +310,18 @@ function withGlobal(_global) {
...
+     * @param {Clock} clock
+     * @param {Timer} job
      */
     function getInfiniteLoopError(clock, job) {
         const infiniteLoopError = new Error(
             `Aborting after running ${clock.loopLimit} timers, assuming an infinite loop!`
         );
 
+        if (!job.error) {
+            return infiniteLoopError
```
benjamingr, line 322: semicolon?

benjamingr (approved):

src/fake-timers-src.js
SimenB: hah, no CI means reviewer needs to notice lint errors 🙈

SimenB: wait, there's no lint rule for semis 🤦

fatso83:
So ... what's the order of things here? #398 and then #401 and then #402?

SimenB:
I think this, release, then figure out what to do in #398 (I'm not totally happy with it)

SimenB:
I'll rebase both to trigger CI 🙂

PR #402, opened 2021-09-22
chore: enforce semicolons via ESLint

SimenB:
#### Purpose (TL;DR) - mandatory

See https://github.com/sinonjs/fake-timers/pull/401#discussion_r713692379

PR #404, opened 2021-09-22
chore: run prettier as part of eslint

SimenB:

I didn't know we had prettier set up (https://github.com/sinonjs/fake-timers/runs/3673842832?check_suite_focus=true) - it's better to just run it as part of `eslint` both for `eslint --fix` and the pre-commit hook. And my IDE will now autoformat on save 🙂 

Note that we can probably remove the `prettier --check` thing as `eslint` will now report any formatting errors.

Negates the need for #402

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -1,8 +1,6 @@
 extends:
   - "@sinonjs/eslint-config"
-
-rules:
-  semi: error
+  - "plugin:prettier/recommended"
 
 overrides:
   - files: "**/*-test.js"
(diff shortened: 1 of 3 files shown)


fatso83:
Nice cleanup!

mroderick:
I don't think this was necessary, but if it makes you happier to configure your IDE this way ... 🤷

fatso83:
Does all the things the old stuff did and a bit more, while making the dev tools people use play better: seems like a win to me 👍

PR #412, opened 2022-01-26
fix global performance node 16

benjamingr:

Fixes `performance` on Node 16 which has a global `performance` but not a global `Performance` 

cc @fatso83 @SimenB @itayperry

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -1,13 +1,6 @@
 "use strict";
 
-// Node.js stores certain things like the performance object on globalThis
-// but not on its `global` object which means we prefer it to `global` in
-// node versions that support it - diverging from the @sinonjs/commons behavior
-/* global globalThis */
-const globalObject =
-    typeof globalThis !== "undefined"
-        ? globalThis
-        : require("@sinonjs/commons").global;
+const globalObject = require("@sinonjs/commons").global;
 
 /**
  * @typedef {object} IdleDeadline
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -1,6 +1,13 @@
 "use strict";
 
-const globalObject = require("@sinonjs/commons").global;
+// Node.js stores certain things like the performance object on globalThis
+// but not on its `global` object which means we prefer it to `global` in
+// node versions that support it - diverging from the @sinonjs/commons behavior
+/* global globalThis */
+const globalObject =
+    typeof globalThis !== "undefined"
+        ? globalThis
+        : require("@sinonjs/commons").global;
 
 /**
  * @typedef {object} IdleDeadline
@@ -157,6 +164,10 @@ function withGlobal(_global) {
     const hasPerformancePrototype =
         _global.Performance &&
         (typeof _global.Performance).match(/^(function|object)$/);
+    const hasPerformanceConstructorPrototype =
+        _global.performance &&
+        _global.performance.constructor &&
+        _global.performance.constructor.prototype;
     const queueMicrotaskPresent = _global.hasOwnProperty("queueMicrotask");
     const requestAnimationFramePresent =
         _global.requestAnimationFrame &&
@@ -738,6 +749,7 @@ function withGlobal(_global) {
 
     /**
      * Gets clear handler name for a given timer type
+     *
      * @param {string} ttype
      */
     function getClearHandler(ttype) {
@@ -749,6 +761,7 @@ function withGlobal(_global) {
 
     /**
      * Gets schedule handler name for a given timer type
+     *
      * @param {string} ttype
      */
     function getScheduleHandler(ttype) {
@@ -1662,8 +1675,15 @@ function withGlobal(_global) {
         }
 
         if (clock.methods.includes("performance")) {
-            if (hasPerformancePrototype) {
-                var proto = _global.Performance.prototype;
+            const proto = (() => {
+                if (hasPerformancePrototype) {
+                    return _global.Performance.prototype;
+                }
+                if (hasPerformanceConstructorPrototype) {
+                    return _global.performance.constructor.prototype;
+                }
+            })();
+            if (proto) {
                 Object.getOwnPropertyNames(proto).forEach(function (name) {
                     if (name !== "now") {
                         clock.performance[name] =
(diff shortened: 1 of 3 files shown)


src/fake-timers-src.js
```diff
@@ -1,6 +1,13 @@
 "use strict";
 
-const globalObject = require("@sinonjs/commons").global;
+// Node.js stores certain things like the performance object on globalThis
+// but not on its `global` object which means we prefer it to `global` in
+// node versions that support it - diverging from the @sinonjs/commons behavior
+/* global globalThis */
```
fatso83, line 6: This is all fine and dandy, but is there a reason we should not include this change in `@sinonjs/commons`?

fatso83 (commented):
Big <3 for this. Just a question before merging.

src/fake-timers-src.js
benjamingr: Well, it changes the meaning of what `global` is and I am not sure that applies to other sinon projects. Let me take another look at the code.

benjamingr: I'm trying to figure out why performance won't be on `global`

test/fake-timers-test.js
```diff
@@ -3775,11 +3778,26 @@ describe("FakeTimers", function () {
                 delete Performance.prototype.someFunc3;
             });
 
+            it("should mock performance on Node 16+", function () {
+                // node 16+ has a performance object but not a global constructor
+                if (typeof performance === "undefined") {
+                    this.skip();
```
SimenB, line 3784: shouldn't there be a `return` after the `skip`s so we don't do further works that's ignored?

src/fake-timers-src.js
```diff
@@ -1672,7 +1685,7 @@ function withGlobal(_global) {
                                 : NOOP;
                     }
                 });
-            } else if (config.toFake.includes("performance")) {
+            } else if ((config.toFake || []).includes("performance")) {
```
SimenB, line 1688: seems unrelated? when is it missing, and how hasn't that been an issue before?

benjamingr: This has been an issue with Node 16 causing sinon to crash. It's here now in case something in the future reaches this branch

test/fake-timers-test.js
benjamingr: I'm... not sure should there?

src/fake-timers-src.js
itayperry: > how hasn't that been an issue before?

It was in the latest merged PR #400 and it works on Node 14-15 but not on Node 16  (it was my mistake)

see https://github.com/sinonjs/fake-timers/issues/394#issuecomment-1013758365

SimenB: cool. can we add node 16 (and 17) to CI with this PR?

test/fake-timers-test.js
benjamingr: No other places in the code do this - so I guessing "no" but I'm just cargo-culting here

src/fake-timers-src.js
benjamingr: Sure, added Node 16

benjamingr: Let me remove 10 which is EoL and has stopped receiving security updates and add Node 17 and a test for something that exists there

benjamingr: Node 17 has unrelated failures, let me try and fix those

benjamingr: I'll also add a bunch of todos for stuff that doesn't work atm

fatso83: Not totally surehow to get that coverage thing going.

SimenB: removing 10 is a breaking change, fwiw

fatso83: @SimenB Nothing is broken and what we actually run the tests on is not our compatibility test. We have an official policy for all Sinon projects: https://github.com/sinonjs/sinon/blob/master/COMPATIBILITY.md#runtimes which basically says "All supported LTS versions". Once 10 no longer is supported, then it's off the table. No need to bump major version numbers. Not that it matters all that much :smiley:

SimenB: Right, I just saw "Let me remove 10 which is EoL". If that doesn't happen then my comment can be ignored 👍 

---

It's still a breaking change to remove remove support for a runtime, which warrants a new major version regardless of what some markdown file says (unless it also says "we do not follow semver, and a new major means incompatible API change" or some such).

Anyways, this probably got a bit theoretical and not related to the PR at hand 😀

SimenB: > (I don't think supporting EoL versions of Node makes sense since they have bugs/security issues that are never going to be fixed)

I don't disagree with that statement. I'm not saying `@sinonjs/fake-timers` should keep supporting EOL versions of Node indefinitely (or even that this PR shouldn't remove support for it, although that seems orthogonal to fixing support for the current LTS release), but _if_ (when) support is dropped, that's a breaking change which should trigger a new major version.

benjamingr:
Let's merge and open another PR to discuss versioning?

fatso83:
I am merging and ignoring the codecov failures due to dropped coverage, but it would be good if that was analyzed somehow so that we understand why. I am guessing the various "if x is present, run this, else that" is only covered if the tests are run in several runtimes (which it is), but are the codecov results just one of these runs or the aggregated result? If the former, then it's hard to do something about it, and if the latter, it is weird that we get a drop.

test/fake-timers-test.js
fatso83: Either `return` or use a `if/else`. Ref the [Mocha docs:](https://mochajs.org/#:~:text=Best%20practice%3A%20To%20avoid%20confusion%2C%20do%20not%20execute%20further%20instructions%20in%20a%20test%20or%20hook%20after%20calling%20this.skip().)
> _Best practice:_ To avoid confusion, do not execute further instructions in a test or hook after calling this.skip().

I can fix this.

jakub-g:
FWIW for future readers: using fake-timers@9.1.2 with nodejs 16.3.0:

```js
jest.useFakeTimers({
    now: ...
});
```

I have 

```
TypeError: Cannot assign to read only property 'performance' of object '[object global]'
```

Solution: upgrade to nodejs 16.5.0+ or use

```js
jest.useFakeTimers({
    now: ...
    doNotFake: ['performance'],
});
```

PR #417, opened 2022-01-28
chore(ci): do not hard code node versions

SimenB:

I assume the purpose is to always run on an LTS version of node here. So specifying `lts/*` does the same thing, but does not have to be updated when v18 (and v20 etc) becomes LTS

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
             - uses: actions/checkout@v2
             - uses: actions/setup-node@v2
               with:
-                  node-version: "16"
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -91,7 +91,7 @@ jobs:
             - uses: actions/checkout@v2
             - uses: actions/setup-node@v2
               with:
-                  node-version: "14"
```
SimenB, line 94: case in point 🙂 

(unless I'm missing something and this is 14 for a reason)

fatso83:
Great small fix!

Issue #358, opened 2021-01-19
requestIdleCallback: callback's IdleDeadline API is not implemented


dvoytenko:
> We understand you have a problem and are in a hurry, but please provide us with some info to make it much more likely for your issue to be understood, worked on and resolved quickly.

* FakeTimers version : appears to still exist at HEAD
* Environment   : any
* Other libraries you are using: sinon

**What did you expect to happen?**

`install()` by default patches `requestIdleCallback`. But when the callback is called, it doesn't have `IdleDeadline` argument. It's described [here](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback). Thus a code that may use this argument will fail.

**How to reproduce**
> Describe *with code* how to reproduce the faulty behaviour
> or link to code on JSBin or similar

A sample code is:

```
requestIdleCallback(function(deadline) {
  if (deadline.timeRemaining() > 0) {
    ...
  }
})
```

fatso83:
Thanks for noting this!

P.S. Quickest way of getting your itch scratch is scratching it yourselves. PRs welcome ;)

fatso83:
Hi, I looked into this, and found several issues.
1. Your issue - the callback not getting the argument.
2. The actual current implementation is wrong. It uses a `number` for the timeout, instad of `{ timeout }` (an object).
2. It does nothing with the timeout. It should schedule the timer to be called at `timeout` if it has not been called.

fatso83:
This is actually mostly done in #421, but requires some debugging to fix the broken tests

PR #421, opened 2022-02-13
IdleDeadline feature

soferdani:
Fix issue #358 create IdleDeadline functionality for the requestIdleCallback 

Please have a look at the requestIdleCallback test section I comment out. I couldn't make this test work, and I think it's because the new feature open the callback and call it with the IdleDeadline object I created.


fatso83:
Thanks for this. I did have a quick look, but I guess getting those stack traces right is non-trivial and I can't see exactly what it is to be done by a quick glance.

fatso83:
It's been two years and it has not moved much. I have no personal interest in this, so for this to go forward, someone needs to pick it up. I can't debug on something I really do not know anything about. Will close in near future.

Issue #422, opened 2022-02-13
When `shouldAdvanceTime: true` and doing some synchronous work, time should still pass


rluvaton:
> We understand you have a problem and are in a hurry, but please provide us with some info to make it much more likely for your issue to be understood, worked on and resolved quickly.

-   FakeTimers version : 9.1.0
-   Environment : macOS Moterey 12.2.1
-   Example URL : 
-   Other libraries you are using: `tap`

**What did you expect to happen?**

Time should still pass even when doing synchronous work

**What actually happens**
time is not updating

**How to reproduce**

Added test case in #423

rluvaton:
Never mind:
> Please note that this is achieved using the original `setImmediate()` API

It stated at the docs how it's implemented which means `setImmediate` won't be called when sync code is running because of how the event loop works...

PR #423, opened 2022-02-13
test(shouldAdvanceTime): add test that check that time is still updated even with sync work

rluvaton:

Adding test case for #422

### DIFF
--- a/test/fake-timers-test.js
+++ b/test/fake-timers-test.js
@@ -4026,6 +4026,24 @@ describe("FakeTimers", function () {
                 done();
             }, 0);
         });
+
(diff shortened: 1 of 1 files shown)


rluvaton:
@benjamingr

benjamingr (approved):

rluvaton:
I think you should stop the tests this will just run until killed, the tests are stuck because of the sync code

benjamingr:
Ok, talked about how to fix it by waiting for a timer on whatsapp

rluvaton:
Closing, see https://github.com/sinonjs/fake-timers/issues/422#issuecomment-1038403540 why

PR #425, opened 2022-03-02
fix(timer.refresh): should just change `callAt`

SimenB:

`timer.refresh` looks completely broken to me (and is definitely at least _somewhat_ broken, see https://github.com/facebook/jest/issues/12527).

1. It uses global `setTimeout` and `clearTimeout`, instead of the ones defined in the `global` passed in
1. it (tries to) remove the old timer and schedule a new one, "changing" its ID. This:
    1. Does not remove the existing one, as that's not an actual real timer
    1. Break identity since we "schedule" (actually, for real, schedules) a _new_ timer instead of refreshing the existing one (try `timer.refresh` in node - `Symbol`s `asyncId` and `triggerId` never changes, only `_idleStart`. also `t.refresh() === t`)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3575,6 +3575,7 @@
       "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-1.2.13.tgz",
       "integrity": "sha512-oWb1Z6mkHIskLzEJ/XWX0srkpkTQ7vaopMQkyaEIoq0fmtFVxOthb8cCxeT+p3ynTdkk/RZwbgG4brR5BeWECw==",
       "dev": true,
+      "hasInstallScript": true,
       "optional": true,
       "os": [
         "darwin"
--- a/test/fake-timers-test.js
+++ b/test/fake-timers-test.js
@@ -4862,6 +4862,23 @@ describe("#187 - Support timeout.refresh in node environments", function () {
         }
         clock.uninstall();
     });
+
+    it("only calls stub once if not fired at time of refresh", function () {
+        const clock = FakeTimers.install();
+        const stub = sinon.stub();
+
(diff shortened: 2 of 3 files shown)


package-lock.json
```diff
@@ -3575,6 +3575,7 @@
       "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-1.2.13.tgz",
       "integrity": "sha512-oWb1Z6mkHIskLzEJ/XWX0srkpkTQ7vaopMQkyaEIoq0fmtFVxOthb8cCxeT+p3ynTdkk/RZwbgG4brR5BeWECw==",
       "dev": true,
+      "hasInstallScript": true,
```
SimenB, line 3578: I just ran `npm install`

test/fake-timers-test.js
```diff
@@ -4862,17 +4862,6 @@ describe("#187 - Support timeout.refresh in node environments", function () {
         }
         clock.uninstall();
     });
-
-    it("assigns a new id to the refreshed timer", function () {
```
SimenB, line 4866: this is plain wrong, as mentioned in the OP

benjamingr (approved):

fatso83:
if good, merge

SimenB:
I'm getting `Error: Could not find browser revision 818858. Run "PUPPETEER_PRODUCT=firefox npm install" or "PUPPETEER_PRODUCT=firefox yarn install" to download a supported Firefox browser binary.` 🙁

SimenB:
Ah, needed to use an older version of npm (which then didn't support the lockfile version...). Then the `postversion` failed, probably because my fork is still named `lolex`. However, all the browser tests ran successfully, so hopefully I didn't mess anything up! 😀 

image

PR #426, opened 2022-04-09
Prohibit faking of faked timers

cjbarth:
#### Purpose

This change will prohibit the faking of already faked timers.

This addresses https://github.com/sinonjs/sinon/issues/2449.

The problem this PR is addressing can occur when an exception is thrown before a timer is `uninstall`ed or if a nested test tries to fake timers again instead of adjusting the currently faked timer or clearing the previous fake first.

If a faked timer is faked again, it will be impossible to recover the original `Date` and `Date`-related functions again in call cases. So, instead of trying to detect incepted faked timers, this is prohibited in all cases.

#### Solution

If an attempt is make to fake a timer that is already faked, an exception will be thrown.

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const globalObject = require("@sinonjs/commons").global;
+const globalDate = globalObject.Date;
 
 /**
  * @typedef {object} IdleDeadline
@@ -133,6 +134,7 @@ const globalObject = require("@sinonjs/commons").global;
  * @returns {FakeTimers}
  */
 function withGlobal(_global) {
+    const stockDate = _global.Date;
     const userAgent = _global.navigator && _global.navigator.userAgent;
     const isRunningInIE = userAgent && userAgent.indexOf("MSIE ") > -1;
     const maxTimeout = Math.pow(2, 31) - 1; //see https://heycam.github.io/webidl/#abstract-opdef-converttoint
@@ -1633,6 +1635,16 @@ function withGlobal(_global) {
                     config
                 )} install requires an object parameter`
             );
+        }
+
+        if (
+            typeof stockDate === "function" &&
+            Math.abs(new Date().getTime() - new stockDate().getTime()) > 100
+        ) {
+            // Allow 100 ms variance for determining already faked timers.
+            // Timers are already faked; this is a problem.
+            // Make the user reset timers before continuing.
+            throw new TypeError("Can't fake timers twice.");
         }
 
         // eslint-disable-next-line no-param-reassign
--- a/test/fake-timers-test.js
+++ b/test/fake-timers-test.js
@@ -53,6 +53,69 @@ const utilPromisifyAvailable = promisePresent && utilPromisify;
 const timeoutResult = global.setTimeout(NOOP, 0);
 const addTimerReturnsObject = typeof timeoutResult === "object";
 
+describe("issue #2449: perminent loss of native functions", () => {
+    it("should not fake faked timers", () => {
+        const currentTime = new Date().getTime();
+        const date1 = new Date("2015-09-25");
+        const date2 = new Date("2015-09-26");
+        let clock = FakeTimers.install({ now: date1 });
+        assert.same(clock.now, date1.getTime());
+        assert.same(new Date().getTime(), 1443139200000);
+        assert.exception(function () {
+            FakeTimers.install({ now: date2 });
+        });
+        clock.uninstall();
+        clock = FakeTimers.install({ now: date2 });
+        assert.same(clock.now, date2.getTime());
+        clock.uninstall();
+        assert.greater(new Date().getTime(), currentTime, true);
+    });
+
+    it("should not fake faked timers on a custom target", () => {
+        const setTimeoutFake = sinon.fake();
+        const context = {
+            Date: Date,
+            setTimeout: setTimeoutFake,
+            clearTimeout: sinon.fake(),
+        };
+        let clock = FakeTimers.withGlobal(context).install();
+        assert.equals(setTimeoutFake.callCount, 1);
+        clock.setTimeout(NOOP, 0);
+        assert.equals(setTimeoutFake.callCount, 1);
+        assert.exception(function () {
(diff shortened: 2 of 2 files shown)


benjamingr:
I'm not sure we should do this - I can see someone writing a test using fake timers inside a suite using fake timers and expecting double faking + restoring to work

cjbarth:
Well, it doesn't now, it fails silently. So, this change will let them know that what they are doing doesn't work. It is actually my guess that if people are trying to get this to work, and their tests are passing (like those in the project I'm working on), then they would probably like to know that their tests are faulty (like I would have wanted to know). It was easy enough to change my tests to be compliant once I knew the problem.

Additionally, no documentation says you can fake faked timers, so if they are doing that, it is undocumented _and_ broken.

test/fake-timers-test.js
```diff
@@ -53,6 +53,69 @@ const utilPromisifyAvailable = promisePresent && utilPromisify;
 const timeoutResult = global.setTimeout(NOOP, 0);
 const addTimerReturnsObject = typeof timeoutResult === "object";
 
+describe("issue #2449: perminent loss of native functions", () => {
```
fatso83, line 56:
```suggestion
describe("issue #2449: permanent loss of native functions", function () {
```
typo-fix. Also Mocha relies heavily on `this`, so it's a malpractice to use arrow functions. Not sure why we do not have ESLint rules to handle this 🤔

src/fake-timers-src.js
```diff
@@ -1635,6 +1637,16 @@ function withGlobal(_global) {
             );
         }
 
+        if (
+            typeof stockDate === "function" &&
+            Math.abs(new Date().getTime() - new stockDate().getTime()) > 100
+        ) {
```
fatso83, line 1643: I do not fancy the implicitness of this. It also won't cover a bunch of scenarios where we _do not_ fake `Date`. fake-timers support faking a single timer or many the APIs. See the `toFake` config option: https://github.com/sinonjs/fake-timers#var-clock--faketimersinstallconfig

I would much rather be specific in testing for a property on the various timers, along the lines of what [Luxon `DateTime.isDatetime(obj)`](https://moment.github.io/luxon/api-docs/index.html#datetime) is doing. This would only mean adding a boolean `isFakeTimer` to Date and the timers. And the test would be explicit!

fatso83 (commented):
I like the _idea_ you propose here, as it seems to support the most common scenario one would want to avoid. Theoretically, someone could be doing the suggested double-faking, but as you pointed out, that does not actually _work_, so if that is the case, I have no objections towards doing this. Just the implementation 🤓

test/fake-timers-test.js
cjbarth: Done. Also, I forgot to run `npm run lint`, so I didn't see the problem with the arrow functions. However, since I'm not using `this` in my tests, it wouldn't have mattered.

src/fake-timers-src.js
cjbarth: I noted that the faked timers can't be used unless `install` is first called, so I tried to put the property on the global only when `install` is called. I tried to leave the global object clean after `uninstall`.

cjbarth:
This global leak detection thing is really causing a problem. It seems that adding a property, even if I remove it when done, prohibits Mocha leak detection from working correctly with `beforeEach` and `afterEach`. Thoughts?

mantoni:
Instead of leaking a global variable, you could take a reference of `setTimeout` and check if it isn't the original instance anymore, no?

mroderick:
This is a very nice pull request, as it will help people where we are failing them 👍

cjbarth:
Ok, I've made another adjustment that is actually even more correct than my first try at this. This tracks very well if the object is a fake or not.

benjamingr (approved):
Changes lgtm, good tests :)

cjbarth:
@benjamingr , can you please run the workflow too?

cjbarth:
Does anyone have any idea why all the tests would run fine on my machine, and apparently here, but not on `chromium`? I don't quite follow what the test that failed is trying to do, and since it runs on my machine, I don't have a good way to trace down what is going on. Honestly, it seems like this might actually be a problem with the test trying to `install` twice. Or `uninstall`ing improperly in `chromium`.

benjamingr:
It looks like most tests pass and there is a single related failure at:

```
 1) loop limit stack trace
       "before each" hook for "provides a stack trace for running all sync":
     TypeError: Can't install fake timers twice on the same global object.
      at Object.install (src/fake-timers-src.js:1642)
      at Context.<anonymous> (test/fake-timers-test.js:5008)
```

cjbarth:
So, after further research, I found that this failing test is actually leaking a faked date instance. So, even the tests for the project were in need of this exception to catch bad use of these timers! I'll push up a fix in a bit.

cjbarth:
Actually, I'm not sure how I might fix this. The problem is that you are skipping a test, in which case, the `afterEach` of the parent `describe` doesn't run because technically the test never ran. That leaks the fake timer. If I move the `install`ation of the fake timer to each lower-level `describe` block, it breaks some of the recursive stuff you're trying to do. Thoughts?

fatso83:
Thank you for sticking with us all the way through. Great work and great tests, and that you discovered a bug in the test suite just shows all the more so how this prevents common cases 🙌

test/fake-timers-test.js
```diff
@@ -53,6 +53,81 @@ const utilPromisifyAvailable = promisePresent && utilPromisify;
...
+            clock = FakeTimers.withGlobal(context).install();
+        });
+        clock.uninstall();
+
+        // After uninstaling we should be able to install without issue
+        clock = FakeTimers.withGlobal(context).install();
+        clock.uninstall();
+    });
+
+    it("should not allow a fake on a custom target if the global is faked adn the context inherited from the global", function () {
```
fatso83, line 92:
```suggestion
    it("should not allow a fake on a custom target if the global is faked and the context inherited from the global", function () {
```

fatso83 (approved):

mroderick:
The changes in this PR are published as `@sinonjs/fake-timers@9.1.2` and are included in `sinon@13.0.2`

PR #431, opened 2022-06-14
Add a bit more structure to the tests

mansona:
#### Purpose (TL;DR)

This Change just splits existing tests into their own files to make it a bit easier to manage (rather than one giant file with all the tests in it). I was motivated to do this when I wanted to create a failing test related to potential bug that I might have found and I found it difficult to figure out where to put my test. 

I only moved the tests related to bug reports (i.e. they referenced an issue number) for now so we can decide if we're ok with this approach before splitting the main test file up any more 👍


fatso83:
rebased and fixed flaky test

Issue #432, opened 2022-06-14
Event timeStamp not using fake-timers


mansona:
> We understand you have a problem and are in a hurry, but please provide us with some info to make it much more likely for your issue to be understood, worked on and resolved quickly.

-   FakeTimers version : latest
-   Environment : Browser
-   Example URL : https://github.com/sinonjs/fake-timers/pull/433
-   Other libraries you are using: none

**What did you expect to happen?**

I'm trying to create 2 events with a certain time between them: 

```js
    window.dispatchEvent(new KeyboardEvent('keypress', { key: 'A' }));
    this.clock.tick(70);
    window.dispatchEvent(new KeyboardEvent('keypress', { key: 'B' }));
```
and then if I checked the difference of those timestamps I was expecting that they be exactly 70ms apart:

```js
const eventDiff = secondEvent.timeStamp - firstEvent.timeStamp;
assert.equals(eventDiff, 150);
```

**What actually happens**

The Events seem to just ignore the fake-timer implementation.

**How to reproduce**

See https://github.com/sinonjs/fake-timers/pull/433

fatso83:
Oooh, I wonder if we are able to fix that. If they maintain a reference to an internal clock, the answer is no.

mansona:
well that's exactly why I asked here @fatso83 😂 I figured someone might be able to shed some light on it. I was trying to find some details on how that timeStamp is generated online but I kinda don't know what I'm looking for and didn't get very far 😞

fatso83:
I would say that faking UI event code in the browser is actually not a core feature of `fake-timers`, so not sure we should spend time implement it, but I think this is a pretty valid use case we could at least document how one could implement. I had never heard of it, so I researched this a bit on MDN and it seems that KeyboardEvent < UIEvent < Event and the [`timeStamp` comes from the top of the hierarchy: the Event class](https://developer.mozilla.org/en-US/docs/Web/API/Event/timeStamp).

> This value is the number of milliseconds elapsed from the beginning of the time origin until the event was created. If the global object is [Window](https://developer.mozilla.org/en-US/docs/Web/API/Window), the time origin is the moment the user clicked on the link, or the script that initiated the loading of the document

That means we would just need to replace the`timeStamp` field is with a something that returns something we want, like the current clock time in ms. When playing in DevTools right now on this page, I could see this working fine:

```javascript
// Set up something to capture events for us to play with in DevTools console
[...]

PR #433, opened 2022-06-14
add failing test for issue 432

mansona:
#### Purpose (TL;DR) 

This shows a failing test for the scenario described in #432


fatso83:
As mentioned in https://github.com/sinonjs/fake-timers/issues/432#issuecomment-1159173536, I think dealing with the DOM APIs falls a bit outside the core of what fake-timers should try to do, especially given that there is an easy workaround. 

Thanks anyway for taking the time to document what was bugging you. That makes looking into it much nicer.

PR #440, opened 2022-11-02
Upgrade dependencies

mroderick:

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,5 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+$(npm bin)/lint-staged
+
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "mocha": "10.1.0",
     "mochify": "9.2.0",
     "nyc": "15.1.0",
-    "prettier": "2.2.1"
+    "prettier": "2.7.1"
   },
   "main": "./src/fake-timers-src.js",
   "husky": {
--- a/package.json
+++ b/package.json
@@ -37,8 +37,6 @@
   "devDependencies": {
     "@sinonjs/eslint-config": "4.0.6",
     "@sinonjs/referee-sinon": "10.1.0",
-    "eslint-config-prettier": "8.3.0",
-    "eslint-plugin-prettier": "3.4.1",
     "husky": "4.2.1",
     "jsdom": "16.5.2",
     "lint-staged": "10.0.7",
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "husky": "4.2.1",
     "jsdom": "16.5.2",
     "lint-staged": "10.0.7",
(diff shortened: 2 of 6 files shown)


fatso83:
Since rebasing and doing all the upgrades, Node 14 complains:
```
  npm ci
npm ERR! fsevents not accessible from chokidar
```

mroderick:
> Since rebasing and doing all the upgrades, Node 14 complains

There's always one squeaky wheel, isn't there? :P

mroderick:
Thank you ❤️

package.json
```diff
@@ -50,6 +50,9 @@
   "dependencies": {
     "@sinonjs/commons": "^2.0.0"
   },
+  "optionalDependencies": {
```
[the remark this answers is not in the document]
mroderick: @SimenB that's a fair point. Would you mind making a ticket for it, so we don't forget?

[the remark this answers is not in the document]
fatso83: AFAIK when looking at this, it was `chokidar`. And this was only an issue with Node 14/NPM 6, not with any later version, so I was thinking it was more of an issue of it only being able to handle the v1 of the package-lock format, whereas the package-lock file has v2.

And I think we _want_ all consumers of mac to install this? This package ensures it uses file system events instead of polling.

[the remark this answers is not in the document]
SimenB: > AFAIK when looking at this, it was `chokidar`. And this was only an issue with Node 14/NPM 6, not with any later version, so I was thinking it was more of an issue of it only being able to handle the v1 of the package-lock format, whereas the package-lock file has v2.

Could we upgrade the version of npm used on CI so the lockfile is supported? Or stay on v1 I guess (unless newer npm just updates the format automatically)

> And I think we _want_ all consumers of mac to install this? This package ensures it uses file system events instead of polling.

`@sinonjs/fake-timers` doesn't do any FS at all, tho, so it shouldn't come with `fsevents` (or `chokidar`).

.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
```
SimenB, line 2: does this work for you? I'm getting

image

when I try to commit

SimenB: `npm bin` fails for me

image

using `npx` works - I can send a PR

SimenB: #457

package.json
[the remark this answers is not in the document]
SimenB: #455

.husky/pre-commit
fatso83: whaaat. What kind of NPM are you running? `npm bin` works even if `alias npm=pnpm`.

fatso83: I need to test this locally.

fatso83: image

Works fine? Is not `npx` potentially just a lot slower?

PR #442, opened 2022-11-07
Prefer @sinonjs/commons@2

mroderick:
That makes ES2017 support explicit

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
   },
   "main": "./src/fake-timers-src.js",
   "dependencies": {
-    "@sinonjs/commons": "^1.7.0"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -49,7 +49,7 @@
   },
   "main": "./src/fake-timers-src.js",
   "dependencies": {
-    "@sinonjs/commons": "^1.7.0"
+    "@sinonjs/commons": "2.0.0"
```
SimenB, line 52: why not semver range?

mroderick: That's a mistake on my part. I'll fix that in `main`

mroderick: https://github.com/sinonjs/fake-timers/pull/447

Issue #441, opened 2022-11-02
Test failing since Node 18.8.0


mroderick:
While working on #439 and #440 with @hexeberlin, we discovered that one unit test is failing since Node 18.8.0

```
  1) FakeTimers
       stubTimers
         should replace the getEntries, getEntriesByX methods with noops that return []:
     TypeError: Cannot assign to read only property 'getEntries' of object '#<Performance>'
      at Context.<anonymous> (test/fake-timers-test.js:3884:50)
      at process.processImmediate (node:internal/timers:471:21)
```

#### How to reproduce

1. `nmv install 18.7`
2. `npm run test-node`
3. Observe all tests pass
1. `nvm install 18.8`
4. `npm run test-node`
5. Observe one test fails

Since this library is downloaded more ~19M times/month and Node 18 has been out for about a month and we have yet to see any error reports about failures in newer Node 18 versions, I think it's safe to say that it's only the test that is failing.

We should fix this soon, as the failing test is preventing us from merging pull requests (or alternatively, we should lock the Node 18 version to 18.7, while we fix this issue).

Plot twist: the tests pass fine in Node v19.0.0

fatso83:
Looking into this. Seems to be some kind of weird runtime issue:
![image](https://user-images.githubusercontent.com/618076/203041065-ad2eb760-9572-4df7-9baf-bb6a8298d75d.png)

Essentially, if _I_ configure the property descriptor in a debugger with the line from the code, it works. If it is just Node running on its own, it has no effect and fails.

```javascript
Object.defineProperty(Performance.prototype, "getEntries", { writable: true })
```

fatso83:
Ha, no, this is just about spelling! I corrected the spelling when doing this. The prop is called `writable`, but we have used `writeable` :rofl:

PR #451, opened 2022-11-19
Fix(runToLastAsync): Prevent infinite loop

### LINKED ISSUE #450: `runToLastAsync` doing infinite loop when `now` is specified to setup clock
While working on https://github.com/vitest-dev/vitest/pull/2209 I have an issue with `runToLastAsync` doing an infinite loop
Below you can find a minimal test to reproduce the issue

-   FakeTimers version : 9.1.2
-   Environment : MACOS
-   Example URL : a minimial reproduction test

-   Other libraries you are using: none

An infinite loop shouldn't be possible and this test shouldn't break

**What actually happens**
This test times out

**How to reproduce**
Here is a minimal test to reproduce the issue

```typescript
        it("new timers created from promises cannot cause an infinite loop when a date is given", function () {
            this.clock = FakeTimers.createClock(Date.now());
            const test = this;
            const spy = sinon.spy();
            const recursiveCallback = function () {
                global.Promise.resolve().then(function () {
                    test.clock.setTimeout(recursiveCallback, 0);
                });
            };

            this.clock.setTimeout(recursiveCallback, 0);
            this.clock.setTimeout(spy, 100);

            return this.clock.runToLastAsync().then(function () {
                assert.isTrue(spy.called);
            });
        });
```

The error for this test is

Capture d’écran 2022-11-19 à 01 08 23

guillaumeduboc:
- feat: add test for issue #450
- fix(runToLastAsync): tick the correct amount of time

Fix #450 
In the `runToLastAsync` method, wait the correct amount of time with `tickAsync`

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -1562,7 +1562,7 @@ function withGlobal(_global) {
                                 resolve(clock.now);
                             }
 
-                            resolve(clock.tickAsync(timer.callAt));
+                            resolve(clock.tickAsync(timer.callAt - clock.now));
                         } catch (e) {
                             reject(e);
                         }
--- a/test/fake-timers-test.js
+++ b/test/fake-timers-test.js
@@ -53,6 +53,25 @@ const utilPromisifyAvailable = promisePresent && utilPromisify;
 const timeoutResult = global.setTimeout(NOOP, 0);
 const addTimerReturnsObject = typeof timeoutResult === "object";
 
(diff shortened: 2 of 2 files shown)


benjamingr (approved):

fatso83 (approved):
I love the diff <3 Thanks for taking the time!

fatso83:
Ah, the tests fail on Node 18 ( #441 )
image

PR #457, opened 2022-11-25
chore: use npx in pre-commit hook

SimenB:

See https://github.com/sinonjs/fake-timers/pull/440/files#r1032069294

https://typicode.github.io/husky/#/?id=locally-installed-binaries

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,5 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-$(npm bin)/lint-staged
+npx lint-staged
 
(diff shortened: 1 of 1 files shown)


.husky/pre-commit
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-$(npm bin)/lint-staged
+npx --no lint-staged
```
SimenB, line 4: we could use `npm exec` as well, but `npx` should be installed on all supported versions of node

fatso83: What kind of NPM do you have installed that does not have the `bin` subcommand? What does `npm --version` say if `npm bin` fails?

SimenB: It was removed in npm v9: https://github.com/npm/cli/releases/tag/v9.0.0

fatso83: Ah, OK, that seals the deal for me.

fatso83 (approved):

Issue #454, opened 2022-11-24
requestAnimationFrame is passed incorrect argument


CreativeTechGuy:
-   FakeTimers version : 10.0.0
-   Environment : Jest/JSDOM
-   Example URL : N/A
-   Other libraries you are using: N/A

**What did you expect to happen?**

The callback passed to `requestAnimationFrame` should receive an argument which is the number of seconds since time origin (same as `performance.now()`). [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame#parameters)

**What actually happens**

The callback instead receives `clock.now` which when compared to `performance.now()` is almost always orders of magnitude higher.

**How to reproduce**

```js
// Sometime before:
clock.setSystemTime(new Date(2000, 10, 2));
// Test code
let lastFrameTime = performance.now();
window.requestAnimationFrame((time) => {
    console.log("Delta Time:", time - lastFrameTime); // This will log something like 973152000000
});
```

**Suggestion**

I know that `performance.now()` may not always be available. So I propose that [this line](https://github.com/sinonjs/fake-timers/blob/2f079447912dd07472461ed7a48b49e28c1041bf/src/fake-timers-src.js#L1225) be changed from:

```js
args: [clock.now + getTimeToNextFrame()],
```

to

```js
args: [(performancePresent ? clock.performance.now() : clock.now) + getTimeToNextFrame()],
```

(I am happy to create a PR if you'd like.)

fatso83:
There might be a bug here, but I think the description here is wrong. You say

> The callback passed to requestAnimationFrame should receive an argument which is the number of seconds since time origin

This is not in line with the docs you refer to:

> "The callback function is passed one single argument, a [DOMHighResTimeStamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp) similar to the one returned by [performance.now()](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now), indicating the point in time when requestAnimationFrame() starts to execute callback functions.

A [DOMHighResTimeStamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp) is not "the number of seconds since time origin", but a millisecond value with high precision. 

> The DOMHighResTimeStamp type is a double and is used to store a time value in milliseconds.

[...]

PR #458, opened 2022-12-02
fix: requestAnimationFrame args

CreativeTechGuy:
#### Purpose (TL;DR)

Fix issue #454 by computing requestAnimationFrame args at the time when the callback is fired.


src/fake-timers-src.js
```diff
@@ -740,7 +740,13 @@ function withGlobal(_global) {
         }
 
         if (typeof timer.func === "function") {
-            timer.func.apply(null, timer.args);
+            if (timer.animation) {
```
fatso83, line 743: This branch needs test coverage

fatso83 (changes requested):
Looks fine, but lacking a test to verify its functionality

src/fake-timers-src.js
CreativeTechGuy: Yes of course! Added.

CreativeTechGuy:
For the record, I am preserving the existing behavior which passes `clock.now` in the case where `performance.now()` isn't available. But that should probably be changed too since that doesn't match the browser's behavior. Should we break out the function that is used to implement `performance.now()` to be usable even when that method isn't mocked? That way `requestAnimationFrame`'s args could be accurate in all cases?

fatso83:
> Should we break out the function that is used to implement performance.now() to be usable even when that method isn't mocked?

Yes, I think that makes perfect sense! Separate commit and separate test case :)

CreativeTechGuy:
You got it @fatso83! New commit and updated tests. Hope you like it 😃

src/fake-timers-src.js
```diff
@@ -740,7 +740,12 @@ function withGlobal(_global) {
...
+                // For requestAnimationFrame, the arg is a reference to performance.now
+                timer.func.apply(null, [timer.args[0]()]);
```
fatso83, line 745: This code is kind of hairy without the comment.  I would like to clean it up somewhat.

The easiest option is probably just to extract this into a named function, like "handleRequestAnimationFrame" and inline the comment there, but it doesn't really remove the main issue, which is tightly coupling two separate bits of code (the setup and the execution of the timer). 

It would be better if we could handle this in a single branch.

What about simply moving this to the setup phase, wrapping it into a closure, so that this bit if the code will just be applying the argument to the function, no matter the type?

I am writing on my phone, so maybe o am missing something here, like args being bound too early, but conceptually that should be cleaner. Your tests should be unaffected by this.

src/fake-timers-src.js
```diff
@@ -1222,7 +1233,7 @@ function withGlobal(_global) {
             const result = addTimer(clock, {
                 func: func,
                 delay: getTimeToNextFrame(),
-                args: [clock.now + getTimeToNextFrame()],
+                args: [FakePerformanceNow],
```
fatso83, line 1236: This is where I would create the closure, if possible.

fatso83 (commented):
I'll await what you think about my comments. Otherwise great stuff.

test/fake-timers-test.js
```diff
@@ -3932,10 +3932,31 @@ describe("FakeTimers", function () {
...
+                    now: sinon.fake(),
+                },
+            }).createClock(123456789);
             const stub = sinon.stub();
-            this.clock.requestAnimationFrame(stub);
-            this.clock.tick(16);
+            clock.requestAnimationFrame(stub);
+            clock.tick(100);
+
+            assert(stub.calledWith(16));
```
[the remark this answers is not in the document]
CreativeTechGuy: Maybe this is unnecessary, but the larger tick number was to clearly convey that this will only be called once and the large tick doesn't affect it. I think 20 would also do the trick to prove that the callback arg is different than the tick amount. I'll update it to 20.

src/fake-timers-src.js
CreativeTechGuy: Yeah I didn't like this either... I don't think we can avoid passing the function around, although possibly it can be less magical and more consistent.

I'll try something in `addTimer` to normalize it and then you can tell me what you think.

CreativeTechGuy: Oh here's an interesting idea, I just tried making `args` a getter for `requestAnimationFrame` and that works great. It creates the closure and the usage is identical since the consuming code cannot tell the difference between `timer.args` being an array or a getter which returns an array.

Hopefully it's not too "clever".

src/fake-timers-src.js
CreativeTechGuy: Got it! I think it's better but let me know what you think!

CreativeTechGuy:
@fatso83 Let me know if the latest changes address all of your concerns. I think this is good to go!

SimenB:
Lockfile was changed from v1 to v2, probably not on purpose 🙂

CreativeTechGuy:
@SimenB Oh no! I'm not even sure how that could have accidentally happened. I would have understood if it upgraded to v3 somehow, but downgraded? No clue! I blame GitHub Workspaces' auto-setup haha.

How would you like me to fix it? I definitely didn't intend to change anything in that file, I'm so sorry it got committed.

SimenB:
I wouldn't worry about it, it doesn't really matter. Just a bit of extra diff 👍

CreativeTechGuy:
I tried to reproduce it with a fresh GitHub Codespace and instantly upon setup it converted it back to v2 so I have no clue what I did. I guess whoever does the next PR will likely (either intentionally or accidentally) upgrade it to v2 again! haha

Issue #459, opened 2022-12-06
[Feature]: Speed up all fake timers


Noriller:
#16 was closed, but I feel it would be good to have something like this.

My problem is that many packages/component libraries use timers internally and you can't just fake and run. Especially in integration tests, where multiple things happen, I had tests breaking because I can't finish an action and run the timers (I also don't want to spam running timers everywhere)

I understand that this package is a dependency of jest and for frontend testing, I would like to "warp" the speed 100x~1000x, so microanimations that would normally take 200ms would take 2ms or less, greatly speeding up tests with minimal change (just adding the speed after using the fake timers) and the best part is without having to spam running timers or mocking animations frameworks (that sometimes are deeply nested in the component libraries).

I feel the existing `config.shouldAdvanceTime` is limited since you can only skip x ms by real x ms.
Meanwhile "warp" would simply run everything faster without the throttling that advance time does.

SimenB:
I don't really follow this. Why not use `clock.runAll()` or something? Or pass "warp" to `tick`, like `clock.tick(16 * 200)`. Do you have an example of a test where this feature would allow it to run faster?

Noriller:
The motivation is I'm using ant design at work and this is what their tests look like:

https://github.com/ant-design/ant-design/blob/4.x-stable/components/menu/__tests__/index.test.tsx
https://github.com/ant-design/ant-design/blob/4.x-stable/components/select/__tests__/index.test.tsx

Basically, they sprinkle "runAll" all over, probably because of micro animations.

Those are only unit tests, and there are cases where they need to "runAll" between each new event.

I have the same problem at work, but in some cases, I would need to "runAll" between some events I shouldn't need to, or have to wait and have the tests running more slowly.

If I can have the timers running (like with `shouldAdvanceTime` but with each tick, it moved more than in real-time) then in many instances, all I would need to do was to let it run fake timers with the config and forget.

So, instead of forcing the timers to run, I could think about testing what I want and "compressing" the time it runs. 
(I like to think of: skipping parts in a video vs using a higher playback speed.)

SimenB:
How would you rewrite those tests to use your suggested API?

CreativeTechGuy:
[...]

Issue #16, opened 2015-06-19
Warp speed?


vroy:
I've recently had to speed up timeouts to speed up some tests. I wrote something along the lines of this:

```javascript
oldInterval = window.setInterval
clock = lolex.install(window)
oldInterval(function() {
  clock.tick(1000)
}, 100)
```

Having something like this in lolex could be a nice addition: `clock.engageWarpSpeed(multiple)`

```javascript
clock.engageWarpSpeed(10) // 10 times faster
clock.engageWarpSpeed(2) // 2 times faster
```

Is this something that you consider merging if I open a pull request?

cjohansen:
Hmm, what's the use case for this? If you can speed it up, why not just fake it entirely and trigger the timers yourself?

vroy:
The app has some setTimeouts that I can't control. The idea is that the tests inject this into the app and speed up all setTimeouts equally.

vroy:
It might very well be a niche use case and I understand if you don't want to introduce this to lolex. That's why I asked via an issue before opening a pull request.

cjohansen:
Sure, that's great. Sorry about the late response. It just seems to me that if you can get in there and speed the timer up, you might as well just use fake ones and tick the clock ahead manually?

vroy:
I'll close this for now.

PR #460, opened 2022-12-06
feat: warpFactor

Noriller:

Adding feature: `warpFactor` that changes the normal delay/timeout by the factor provided.
Issues: #459 #16 

This is a dependency of Jest that is widely used. In frontend testing, many component libraries abstract micro animations that take time to complete and slow down testing.

Using fake timers would mean having to spam methods to run the timers, which is not optimal.

This PR aims to add a way to speed up all those hidden timers without anything more than passing a flag with the factor to be used.

Mainly, I added the `warpFactor` to the config and when present, it takes the delay and divides it by the factor.

The one thing I want to highlight is that if the factor is higher than the timeout, then the timeouts are rounded to zero. 

This would throw off the ordering of timers, since you can possibly make all timers 0. So I added something to keep track of what would be the actual `callAt` to make use in the compare timers function.

I'm not sure in what scenario it could be useful, but this implementation also lets you use an fraction as the `warpFactor`, making all timeouts run slower than normal.


benjamingr (commented):
The code itself is good but I don't really understand the feature. I don't understand the motivation to not just `.tick(n * 100)` instead of this?

Also wouldn't it be simpler to just progress `.tick` calls by warpFactor * passedTime instead of modifying the timers themselves?

Noriller:
At the time I didn't think of that.
I screw up this PR, so I'll be opening a new one with the changes.

benjamingr (commented):
Nice the code looks good to me and the implementation looks fine - I am not sure about the feature itself still.

@fatso83 @SimenB wdyt? I think OP makes a compelling case but I'm hesitant approving a feature I don't fully understand the usage of in real life.

fatso83:
Yeah, as Simen asked (to be answered) in the original issue, I also don't understand the usecase or what isn't handled by .runAll() or tick(n*1000).

fatso83:
See  https://github.com/sinonjs/fake-timers/issues/459#issuecomment-1936981297. 

This is not a full stop! It can be reopened if someone provides a good usecase for why this is needed and makes sense!

PR #465, opened 2023-04-28
Add clock.jump method

### LINKED ISSUE #452: [Feature] Jump forward in time to simulate throttled timers

-   FakeTimers version : 10.0.0
-   Environment : Jest/JSDOM
-   Example URL : N/A
-   Other libraries you are using: N/A

I expected there would be an API to enable skipping forward in time in a similar way that [browsers will throttle a `setInterval` when the page is backgrounded](https://developer.chrome.com/blog/timer-throttling-in-chrome-88/). That is to say, if I have the following code:

```js
setInterval(() => {
    console.log("Here");
}, 100);
```

This would print 10 times a second. So running `clock.tick(2000)` would print this 20 times. I expected that there would be a `clock.jump(2000)` which would print this only once, effectively advancing the clock without any intermediary tasks in the event loop being processed. Another way to look at it is that all pending tasks were throttled to whatever value is passed to the `clock.jump` method.

**What actually happens**

There currently is no way to test code which is written to handle this browser behavior of throttling timers.

**How to reproduce**

N/A

CreativeTechGuy:
#### Purpose (TL;DR)

Closes #452 by creating the `clock.jump(time)` method enabling skipping forward in time and running any timers at most once.

### DIFF
--- a/test/fake-timers-test.js
+++ b/test/fake-timers-test.js
@@ -4103,6 +4103,62 @@ describe("FakeTimers", function () {
         });
     });
 
+    describe("jump", function () {
+        beforeEach(function () {
+            this.clock = FakeTimers.install({ now: 0 });
+        });
+
+        afterEach(function () {
+            this.clock.uninstall();
+        });
+
+        it("ignores timers which wouldn't be run", function () {
+            const stub = sinon.stub();
+            this.clock.setTimeout(stub, 1000);
+
+            this.clock.jump(500);
+
+            assert(stub.notCalled);
+        });
+
+        it("pushes back execution time for skipped timers", function () {
+            const stub = sinon.stub();
+            this.clock.setTimeout(stub, 1000);
+
+            this.clock.jump(2000);
+
+            assert(stub.calledOnce);
+        });
+
+        it("handles multiple pending timers and types", function () {
+            const longTimers = [sinon.stub(), sinon.stub()];
+            const shortTimers = [sinon.stub(), sinon.stub(), sinon.stub()];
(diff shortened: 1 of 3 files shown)


test/fake-timers-test.js
```diff
@@ -4103,6 +4103,62 @@ describe("FakeTimers", function () {
...
+            assert(stub.notCalled);
+        });
+
+        it("pushes back execution time for skipped timers", function () {
+            const stub = sinon.stub();
+            this.clock.setTimeout(stub, 1000);
+
+            this.clock.jump(2000);
+
+            assert(stub.calledOnce);
```
amorris-canva, line 4130: Are you also able to test that `stub` is called when the time is `2000` and not when it is `1000` (as it would have been with `clock.tick(2000)`)?

CreativeTechGuy: Good idea! I've updated the test.

fatso83:
I think this looks good. Cannot see any reason not to merge?

fatso83:
Any chance of adding this to the Sinon docs on fake timers? https://github.com/sinonjs/sinon/blob/main/docs/release-source/release/fake-timers.md

fatso83:
This is part of @sinonjs/fake-timers@10.1.0

CreativeTechGuy:
> Any chance of adding this to the Sinon docs on fake timers? https://github.com/sinonjs/sinon/blob/main/docs/release-source/release/fake-timers.md

Yeah. Those docs look far more out of sync than just this, but I can add this new method if you'd like!

Issue #466, opened 2023-05-13
NodeJS "timers" module is not mocked


swenzel-arc:
-   FakeTimers version : v10.1.0
-   Environment : Node.js v18.12.1
-   Other libraries you are using: jest v29.5.0

**What did you expect to happen?**
NodeJS's [timers](https://nodejs.org/api/timers.html) module to be mocked upon `FakeTimers.install()`.

**What actually happens**
It's not mocked.

**How to reproduce**
```js
let timers = require("timers");
let FakeTimers = require("@sinonjs/fake-timers");

let clock;

beforeEach(() => {
    clock = FakeTimers.install();
});

it("doesn't work", () => {
    let timerDone = false;
    timers.setTimeout(()=>{timerDone=true;}, 1000);
    clock.tick(1001);
    expect(timerDone).toBe(true);
});

afterEach(() => {
    clock.uninstall();
});
```

I was testing some code that is working with socket.io and was wondering why my sockets kept closing with ping time outs. Turns out engine.io is using `timers.*` instead of the corresponding global functions.  
TBH I'm not quite sure if a fix belongs into this library or into jest's `useFakeTimers` function. However, the README here says it'd mock all "native" timers when `FakeTimers.install` is called without arguments, so it should maybe also include NodeJS's core module timers.

benjamingr:
PR welcome :)

fatso83:
@benjamingr You are bit more into the Node internals than me, so I just vaguely remember this or something related (HTTP Headers in `http` using primordials?) coming up earlier. Are these timers (or the internal) easily stubbable somehow? Not totally sure where a volunteer would start. Jest would probably have an easier time doing this by just stubbing out calls to the `timers` lib, but we would somehow need to interface with the library directly. I would assume that trying to modify the exports directly would not work (this being ESM, using a immutable namespace and all).

benjamingr:
The timers should be pretty easily stubbable unlike HTTP headers - just `require('timers')` and replace that object's setTimeout/setInterval properties

fatso83:
@swenzel-arc You already have code that works, so that is great. Now what about the interface ... We need to be able to selectively disable or enable the stubbing. I guess it makes sense to enable it as the default? And how should we deal with only stubbing selected timers? I think a reasonable assumption is that if you only want to stub out two specific timers from `global`, you would not want another selection for Node's `timers` module. If that assumption does not hold, we need a way of specifying it.

```javascript
[...]

PR #467, opened 2023-05-16
Add mocks to NodeJS "timers" module

swenzel-arc:

Fix issue #466 

#### Solution
It works like this:

Try to import `timers` module as `timersModule`. If it doesn't work, that's okay.

Upon `FakeTimers.install()`, iff we're installing on the `globalObject` and `timersModule` was successfully imported:
For each property that was mocked on the globalObject iff it also exists on `timersModule`:
Safe its name and original value from `timersModule` to a new property on the Clock object.
Then set it to the same property as the faked property on `globalObject`.

Upon `FakeTimers.uninstall()`, iff the object holding `timersModule`'s originals is truthy:
For each entry set the corresponding property back to the original.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ clock instance, not the browser's internals.
 
 Calling `install` with no arguments achieves this. You can call `uninstall`
 later to restore things as they were again.
+Note that in NodeJS also the [timers](https://nodejs.org/api/timers.html) module will receive fake timers when using global scope.
 
 ```js
 // In the browser distribution, a global `FakeTimers` is already available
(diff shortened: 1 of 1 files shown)


test/issue-466-test.js
```diff
@@ -0,0 +1,144 @@
+"use strict";
```
fatso83, line 1: Stuff the tests inside of this into a suitably named new describe block in `fake-timers-test.js`. We usually reserve the `issues*` files for regression test cases.

test/issue-466-test.js
```diff
@@ -0,0 +1,144 @@
...
+const { sinon } = require("@sinonjs/referee-sinon");
+var timersModule;
+
+if (typeof require === "function" && typeof module === "object") {
+    try {
+        timersModule = require("timers");
+    } catch (e) {
+        // ignored
+    }
+}
```
fatso83, line 13: Just use Mocha's built-ins to selectively run it in supported environments
```javascript
        before(function () {
            if (!timersModule) {
                this.skip();
            }
        });
```
Enables a bit better reporting :)

src/fake-timers-src.js
```diff
@@ -886,6 +894,14 @@ function withGlobal(_global) {
                     }
                 }
             }
+            if (clock.timersModuleMocks) {
```
fatso83, line 897: They are not [actual mocks](https://www.martinfowler.com/articles/mocksArentStubs.html), though, so could trim that bit off.

To be consistent with the `methods` prop, just name it `timersModuleMethods`

fatso83 (changes requested):
I think you could probably afford adding a line to the README that we touch the `"timers"` module, but otherwise this generally looks fine! Just a few minor things.

benjamingr:
> Btw. do you think we should add a config flag that allows to skip altering the timers module?

No but that's a good point, it should respect the regular "should I mock setTimeout/setInterval or whatever" option (toFake)

fatso83 (approved):

fatso83:
Out as 10.2.0

fatso83:
Seems like #470 made a good candidate for why this could have been behind a flag. Some weird regression hitting us.

PR #474, opened 2023-07-18
feat: mock intl api

BenIsenstein:
#### Purpose (TL;DR) - Mock the global Intl API

Solve the breaking of timer-dependant methods on the Intl global object by including a mocked version

Even after calling install(), the Intl.DateTimeFormat() class will currently use the real time of the system, not the time specified by the fake clock.

Affected methods: DateTimeFormat.format(), DateTimeFormat.formatToParts()

My repo demonstrating the problem, using Jest (which uses sinonjs/fake-timers): https://github.com/BenIsenstein/jest_fix_intl_fake_timers

The fix works by following the patterns used to mock the global Date class and mocking the Intl object. A producer function is written to return an Intl object largely identical to the original, except for the DateTimeFormat() class. DateTimeFormat has several methods that are time-sensitive, and depend on the system time when no argument is passed. These are intercepted and made to default to the clock time.

```javascript
//eslint-disable-next-line jsdoc/require-jsdoc
function createIntl() {
    const ClockIntl = { ...NativeIntl };

    ClockIntl.DateTimeFormat = function (...args) {
        const realFormatter = new NativeIntl.DateTimeFormat(...args);
        const formatter = {};

        ["formatRange", "formatRangeToParts", "resolvedOptions"].forEach(
            (method) => {
                formatter[method] =
                    realFormatter[method].bind(realFormatter);
            }
        );

        ["format", "formatToParts"].forEach((method) => {
            formatter[method] = function (date) {
                return realFormatter[method](date || ClockIntl.clock.now);
            };
        });

        return formatter;
    };

    ClockIntl.DateTimeFormat.prototype = Object.create(
        NativeIntl.DateTimeFormat.prototype
    );
    ClockIntl.DateTimeFormat.prototype.constructor =
        ClockIntl.DateTimeFormat;
    ClockIntl.DateTimeFormat.supportedLocalesOf =
        NativeIntl.DateTimeFormat.supportedLocalesOf;

    return ClockIntl;
}
```

This fake Intl object is provided to the Global object when install() is called.

Line 1176, inside createClock()
```javascript
if (intlPresent) {
    clock.Intl = createIntl();
    clock.Intl.clock = clock;
}
```

and the original Intl object is restored during uninstall().

Line 979 inside hijackMethod()
An else-if clause added to handle hijacking Intl. Because Intl isn't a straightforward global function, rather a global object used like a module, it can't be mocked using the default else clause at the end of hijackMethod(). It has its own clause to set the mocked Intl object globally. 
```javascript
else if (method === "Intl") {
```

I'm fairly confident on implementation, however less confident on the tests. All advice is appreciated!

Ben

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 
+11.1.0 / 2023-07-17
+==================
+  * Add a mock for the global Intl object
+
 11.0.0 / 2023-06-12
 ==================
   * Re-release 10.2.0 as a new major version as mocking Node "timers" module broke some setups
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,3 @@
-
-11.1.0 / 2023-07-17
-==================
-  * Add a mock for the global Intl object
-
 11.0.0 / 2023-06-12
 ==================
   * Re-release 10.2.0 as a new major version as mocking Node "timers" module broke some setups
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -206,10 +206,6 @@ function withGlobal(_global) {
         _global.clearImmediate = _global.clearImmediate;
     }
 
-    if (intlPresent) {
-        _global.Intl = _global.Intl;
-    }
-
     /* eslint-enable no-self-assign */
 
     _global.clearTimeout(timeoutResult);
@@ -534,8 +530,7 @@ function withGlobal(_global) {
         ClockIntl.DateTimeFormat.prototype = Object.create(
             NativeIntl.DateTimeFormat.prototype
         );
-        ClockIntl.DateTimeFormat.prototype.constructor =
-            ClockIntl.DateTimeFormat;
+
         ClockIntl.DateTimeFormat.supportedLocalesOf =
             NativeIntl.DateTimeFormat.supportedLocalesOf;
 
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -186,6 +186,7 @@ function withGlobal(_global) {
         typeof _global.cancelIdleCallback === "function";
     const setImmediatePresent =
         _global.setImmediate && typeof _global.setImmediate === "function";
+    const intlPresent = _global.Intl && typeof _global.Intl === "object";
 
     // Make properties writable in IE, as per
     // https://www.adequatelygood.com/Replacing-setTimeout-Globally.html
@@ -204,11 +205,17 @@ function withGlobal(_global) {
         _global.setImmediate = _global.setImmediate;
         _global.clearImmediate = _global.clearImmediate;
     }
+
+    if (intlPresent) {
+        _global.Intl = _global.Intl;
+    }
+
     /* eslint-enable no-self-assign */
 
     _global.clearTimeout(timeoutResult);
 
     const NativeDate = _global.Date;
+    const NativeIntl = _global.Intl;
     let uniqueTimerId = idCounterStart;
 
     /**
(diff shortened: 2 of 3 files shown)


fatso83:
Love when wanted features come with a PR <3 I am off on holiday with just a small mobile screen unsuitable for reviewing code, but hopefully someone else could have a looksie?

benjamingr:
I'm on medical leave I'll try to take a look next week and maybe ask Erick from Node to look

ErickWendel:
As I don't have much context I think it LGTM

BenIsenstein:
Hi all, it looks like the issue in the test suite is being caused by a difference between the mochify and node runtimes. `Intl.DateTimeFormat.resolvedOptions()` returns a different value between the two. Any suggestions on how to resolve this?

fatso83:
It's not mochify (which is running Chrome headless) vs Node, but specifically Node 14 breaking. You see the tests passing in Node 16 and 18. 

If you look at [the engine comparisons page for Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat#:~:text=Toggle%20history-,options.fractionalSecondDigits,-parameter) you can see that `fractionalSecondDigits` was only added in Node 14.6. Guessing we run a slightly older version. 

In any case, we only support Node LTS versions, and Node 14 LTS has reached its EOL, so we can just remove that :)

fatso83:
I have now removed Node 14 from `main`. Does not block this and you do not _need_ to rebase to verify, but know you now 👍 

Do you deem it ready for final review and merge?

BenIsenstein:
Rebased just to keep things clean (Y) yes, it's ready for final review!

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 
+11.1.0 / 2023-07-17
+==================
+  * Add a mock for the global Intl object
+
```
fatso83, line 5: You can usually drop this. This is semi-auto-populated on `npm publish`.

src/fake-timers-src.js
```diff
@@ -204,11 +205,17 @@ function withGlobal(_global) {
         _global.setImmediate = _global.setImmediate;
         _global.clearImmediate = _global.clearImmediate;
     }
+
+    if (intlPresent) {
+        _global.Intl = _global.Intl;
+    }
+
```
fatso83, line 212: Drop this "tautology". The above stuff is only a custom hack for IE, which does not have Intl anyway. I will remove the above, as we no longer support IE.

src/fake-timers-src.js
```diff
@@ -500,6 +507,41 @@ function withGlobal(_global) {
...
+                };
+            });
+
+            return formatter;
+        };
+
+        ClockIntl.DateTimeFormat.prototype = Object.create(
+            NativeIntl.DateTimeFormat.prototype
+        );
+        ClockIntl.DateTimeFormat.prototype.constructor =
```
fatso83, line 537: Could you add a check to verify that the `constructor` property is indeed the faked Intl object and not the native one? Otherwise we might as well drop it, as it's a behavior not specified by tests.

src/fake-timers-src.js
```diff
@@ -500,6 +507,41 @@ function withGlobal(_global) {
...
+            ["formatRange", "formatRangeToParts", "resolvedOptions"].forEach(
+                (method) => {
+                    formatter[method] =
+                        realFormatter[method].bind(realFormatter);
+                }
+            );
+
+            ["format", "formatToParts"].forEach((method) => {
+                formatter[method] = function (date) {
+                    return realFormatter[method](date || ClockIntl.clock.now);
```
fatso83, line 527: From reading the code and checking against the PR description, this is basically the entire reason for this feature (right?): to get correct behaviour (formatting today's date) when leaving [the optional `date` argument out](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatToParts).

fatso83 (changes requested):
This is an excellent PR. Really conforms to everything we want, so thank you very much! I'll merge it once the (tiny) changes have been done.

src/fake-timers-src.js
fatso83: Maybe also just add a basic tests for the supported locales?

src/fake-timers-src.js
BenIsenstein: Yes, that's exactly right! Expose the clock to the `DateTimeFormat` instance, pass the now value as the backup argument.

BenIsenstein:
Thanks for the review! Changes:

- removed changelog entry
- removed IE tautology
- removed setting constructor (setting the prototype takes care of that already, good catch)
- added a test for the `supportedLocalesOf` method

fatso83 (approved):

PR #477, opened 2023-08-07
Remove IE specific hacks

fatso83:

We no longer support Internet Explorer, so remove those hacks.

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -143,8 +143,6 @@ if (typeof require === "function" && typeof module === "object") {
  * @returns {FakeTimers}
  */
 function withGlobal(_global) {
-    const userAgent = _global.navigator && _global.navigator.userAgent;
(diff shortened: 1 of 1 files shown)


src/fake-timers-src.js
```diff
@@ -188,26 +186,6 @@ function withGlobal(_global) {
...
-        _global.clearTimeout = _global.clearTimeout;
-        _global.setInterval = _global.setInterval;
-        _global.clearInterval = _global.clearInterval;
-        _global.Date = _global.Date;
-    }
-
-    // setImmediate is not a standard function
-    // avoid adding the prop to the window object if not present
-    if (setImmediatePresent) {
-        _global.setImmediate = _global.setImmediate;
```
benjamingr, line 205: Node actually has setImmediate as well

fatso83: Yes, not removing that. Just the IE specific x=x thingie.

benjamingr (approved):

PR #479, opened 2023-08-22
feat: add RelativeTimeFormat to Intl API

sebastinez:

PR #474 has introduced a mock for the intl API, and this has introduced a regression for me, where I'm not able to call `Intl.RelativeTimeFormat`.

The problem showed in some of my test cases, once I updated `fake-timers`, when trying to use `Intl.RelativeTimeFormat` I got the following error: `Intl.RelativeTimeFormat is not a constructor`

#### Solution
The solution I opted for is to assign `RelativeTimeFormat` to the mocked `Intl` API with.
```js
ClockIntl.RelativeTimeFormat = NativeIntl.RelativeTimeFormat;
```

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -481,9 +481,24 @@ function withGlobal(_global) {
         return mirrorDateProperties(ClockDate, NativeDate);
     }
 
-    //eslint-disable-next-line jsdoc/require-jsdoc
+    /**
+     * Mirror Intl by default on our fake implementation
+     *
+     * Most of the properties are the original native ones,
+     * but we need to take control of those that have a
+     * dependency on the current clock.
+     *
+     * @returns {object} the partly fake Intl implementation
+     */
     function createIntl() {
-        const ClockIntl = { ...NativeIntl };
+        const ClockIntl = {};
+        /*
+         * All properties of Intl are non-enumerable, so we need
+         * to do a bit of work to get them out.
+         */
+        Object.getOwnPropertyNames(NativeIntl).forEach(
+            (property) => (ClockIntl[property] = NativeIntl[property])
+        );
 
         ClockIntl.DateTimeFormat = function (...args) {
             const realFormatter = new NativeIntl.DateTimeFormat(...args);
@@ -504,8 +519,6 @@ function withGlobal(_global) {
 
             return formatter;
         };
-
-        ClockIntl.RelativeTimeFormat = NativeIntl.RelativeTimeFormat;
 
         ClockIntl.DateTimeFormat.prototype = Object.create(
             NativeIntl.DateTimeFormat.prototype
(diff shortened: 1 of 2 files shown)


fatso83 (approved):
Seems like an OK interim solution.

src/fake-timers-src.js
```diff
@@ -505,6 +505,8 @@ function withGlobal(_global) {
             return formatter;
         };
 
+        ClockIntl.RelativeTimeFormat = NativeIntl.RelativeTimeFormat;
```
SimenB, line 508: Are there others? Could we perhaps loop over the props of `NativeIntl`?

sebastinez: @SimenB yeah there are other methods, and we could loop over them and assign them to `ClockIntl` with something like this.

```js
Object.getOwnPropertyNames(NativeIntl)
    .filter((item) => typeof NativeIntl[item] === "function")
    .forEach((method) => (ClockIntl[method] = NativeIntl[method]));
```

We still would need to keep the new implementation of `DateTimeFormat` due to this LOC
https://github.com/sebastinez/fake-timers/blob/f4469bf39f933b148e400170ae7d698c41079908/src/fake-timers-src.js#L505
Which allows to pass `undefined` to `Intl.DateTimeFormat` here
https://github.com/sebastinez/fake-timers/blob/f4469bf39f933b148e400170ae7d698c41079908/test/fake-timers-test.js#L5193

sebastinez:
Thanks for your reviews, what are the next steps?
Do you need me to sign something or can the PR get merged?

sebastinez:
Hey folks, this issue doesn't allow me to update fake-timers, since it would break my production.
Is there work be done on a better solution? Or is anything holding this PR up?

sebastinez:
Didn't want to close the PR, just comment on it

src/fake-timers-src.js
fatso83: I was not merging due to this seemingly being unresolved .. ?

fatso83: @simenb any objections to his proposal?

SimenB: not at all 🙂

fatso83: PS. The line of code you are referring to is just implementing the Ecmascript spec 
https://tc39.es/ecma402/#sec-datetime-format-functions

SimenB:
Since there's no compilation step in this repo, you can also depend on your fork directly

piotrpalek:
I'm getting the same issue, but with `NumberFormat` :( I think sinon would need to pick up the whole Intl API to fix this

fatso83:
New version is out.

piotrpalek:
Thanks! Sorry for hijacking this PR 😅

Issue #437, opened 2022-08-15
ClockDate class overwrite prototype of instance


Oustinger:
ClockDate class overwrite prototype of instance.
In my real project it happens when I use datepicker library (@eonasdan/tempus-dominus).

-   FakeTimers version : 9.1.2
-   Environment : Opera 89.0.4447.64, Windows 10
-   Example URL : [repository](https://github.com/Oustinger/test_fake-timers_lib)
-   Other libraries you are using: -

**What did you expect to happen?**
I expected to receive instance of DateTime with it's methods

**What actually happens**
Prototype of child instance is not equals to DateTime class, but equals to fake date class

**How to reproduce**
I'd created a [repository](https://github.com/Oustinger/test_fake-timers_lib) with reproducing this problem

i. You can clone and run [project](https://github.com/Oustinger/test_fake-timers_lib)
ii. Or you can create project with @sinonjs/fake-timers and run the code below by yourself

    var FakeTimers = require("@sinonjs/fake-timers");
    
    FakeTimers.install();
    
    class DateTime extends Date {
        constructor() {
            super();
    
            this.bar = 'bar';
        }
    
        foo() {
            console.log('Lorem inpsum');
        }
    }
    
    var dateTime = new DateTime();
    
    dateTime.foo();

benjamingr:
This is probably a bug because when you `new` something in JS and return something different from the constructor - it will return that thing.

```
class Foo {
  constructor() {
     return { x: 5 };
  }
  bar() {} 
}
const f = new Foo();
f.bar(); 
```

We should probably fix the code to not do that _or_ call `Reflect.construct(new.target)` instead of assuming we have a `NativeDate`.

fatso83:
>  when you new something in JS and return something different from the constructor

Where do we do that? I could not see this happening in the example code.

benjamingr:
> Where do we do that? I could not see this happening in the example code.

Our `ClockDate` function doesn't use `this` and instead has a return value even for the case it doesn't return a native date.

fatso83:
@Oustinger The quickest path to seeing a patch commit for this is probably submit a small PR with a regression test. I can make a release quick, but I do not have time atm to debug, fix and verify this.

Oustinger:
@fatso83 I would do that. But I couldn't find a place to fix on my own.

fatso83:
No one would expect you to 🙂 I certainly did not. Do you think you have the hints you need?

Oustinger:
[...]

PR #480, opened 2023-09-05
fix: save methods of children Date instance (#437)

Oustinger:
Fix issue #437 by creating ClockDate as JS class and extending it from NativeDate

Why I did this way: 
* In initinial solution (creating `ClockDate` by JS function) `ClockDate` can't be correct instance to extend from it. That's why I replaced function on class. And it resolved my problem.
* Mechanics of call `Date()` without `new` keyword I solved by using `Proxy` instance.
* Also in my solution removed an `mirrorDateProperties` function. It's not needed any more.

But I had some troubles with tests:
* I skiped [this test](https://github.com/sinonjs/fake-timers/blob/main/test/fake-timers-test.js#L3106-3117), because I didn't understand what it checks and how it should be now
* I skiped [an other test](https://github.com/sinonjs/fake-timers/blob/main/test/fake-timers-test.js#L3224-3226), because I think it's not corresponding for now. `ClockDate.prototype` never equals to `Date.prototype`, because `ClockDate` is extended from `Date`
* Please, check the other tests, that I changed. Hope, I changed them correct

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -398,109 +398,59 @@ function withGlobal(_global) {
         return infiniteLoopError;
     }
 
-    /**
-     * @param {Date} target
-     * @param {Date} source
-     * @returns {Date} the target after modifications
-     */
-    function mirrorDateProperties(target, source) {
-        let prop;
-        for (prop in source) {
-            if (source.hasOwnProperty(prop)) {
-                target[prop] = source[prop];
-            }
-        }
-
-        // set special now implementation
-        if (source.now) {
-            target.now = function now() {
-                return target.clock.now;
-            };
-        } else {
-            delete target.now;
-        }
-
-        // set special toSource implementation
-        if (source.toSource) {
-            target.toSource = function toSource() {
-                return source.toSource();
-            };
-        } else {
-            delete target.toSource;
-        }
-
-        // set special toString implementation
-        target.toString = function toString() {
-            return source.toString();
-        };
-
-        target.prototype = source.prototype;
-        target.parse = source.parse;
-        target.UTC = source.UTC;
-        target.prototype.toUTCString = source.prototype.toUTCString;
-        target.isFake = true;
-
-        return target;
-    }
-
     //eslint-disable-next-line jsdoc/require-jsdoc
     function createDate() {
-        /**
-         * @param {number} year
-         * @param {number} month
-         * @param {number} date
-         * @param {number} hour
-         * @param {number} minute
-         * @param {number} second
-         * @param {number} ms
-         * @returns {Date}
-         */
-        function ClockDate(year, month, date, hour, minute, second, ms) {
-            // the Date constructor called as a function, ref Ecma-262 Edition 5.1, section 15.9.2.
-            // This remains so in the 10th edition of 2019 as well.
-            if (!(this instanceof ClockDate)) {
-                return new NativeDate(ClockDate.clock.now).toString();
-            }
-
-            // if Date is called as a constructor with 'new' keyword
-            // Defensive and verbose to avoid potential harm in passing
-            // explicit undefined when user does not pass argument
-            switch (arguments.length) {
-                case 0:
-                    return new NativeDate(ClockDate.clock.now);
-                case 1:
-                    return new NativeDate(year);
-                case 2:
-                    return new NativeDate(year, month);
-                case 3:
-                    return new NativeDate(year, month, date);
-                case 4:
-                    return new NativeDate(year, month, date, hour);
-                case 5:
-                    return new NativeDate(year, month, date, hour, minute);
-                case 6:
-                    return new NativeDate(
-                        year,
-                        month,
-                        date,
-                        hour,
-                        minute,
-                        second,
-                    );
-                default:
-                    return new NativeDate(
-                        year,
-                        month,
-                        date,
-                        hour,
-                        minute,
-                        second,
-                        ms,
-                    );
+        class ClockDate extends NativeDate {
+            /**
+             * @param {number} year
+             * @param {number} month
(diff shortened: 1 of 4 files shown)


fatso83:
Oooh, I somehow missed that this was added. Sorry about the late look, but while this is interesting, my head is a bit fried 🤯 This is quite a big change, so I (or someone else with a bigger head) needs to take a new look with a fresh head!

Thanks for providing a fix to something that bugged you a year ago. That's persistence we can appreciate 😄

test/issue-437-test.js
```diff
@@ -0,0 +1,29 @@
+"use strict";
+
+const { FakeTimers, assert } = require("./helpers/setup-tests");
+
+describe("issue #437", function () {
+    it("should save methods of children instance", function () {
```
fatso83, line 6:
```suggestion
    it("should save methods of subclass instance", function () {
```

fatso83: Not sure why you talk about child instances? Children of what, if so. This seems like a normal subclass relationship.

test/fake-timers-test.js
```diff
@@ -3283,7 +3283,7 @@ describe("FakeTimers", function () {
         });
 
         it("mirrors toString", function () {
-            assert.same(this.clock.Date.toString(), Date.toString());
```
fatso83, line 3286: I am not sure why this was needed? Do they no longer return the same thing? If so, it's a bit strange that they are the same.

test/fake-timers-test.js
```diff
@@ -3221,7 +3221,7 @@ describe("FakeTimers", function () {
             assert.equals(fakeDateStr, new this.clock.Date().toString());
         });
 
-        it("mirrors native Date.prototype", function () {
+        it.skip("mirrors native Date.prototype", function () {
```
fatso83, line 3224: I guess we could change this to be a test that it passes an instanceof check. `it("is a Date instance", ...`?

test/fake-timers-test.js
```diff
@@ -3103,7 +3103,7 @@ describe("FakeTimers", function () {
             assert(typeof date === "string");
         });
 
-        it("creates real Date objects when Date constructor is gone", function () {
+        it.skip("creates real Date objects when Date constructor is gone", function () {
```
fatso83, line 3106: This test does not seem very useful. I would check the Git history for references.

fatso83 (changes requested):
It's a bit scary breaking change, so I think I would like some more 👀 on this before going ahead.

benjamingr:
> It's a bit scary breaking change, so I think I would like some more 👀 on this before going ahead.

I started reviewing it and came to the "bit scary" conclusion as well.

src/fake-timers-src.js
```diff
@@ -376,109 +376,59 @@ function withGlobal(_global) {
...
+             * @param {number} ms
+             * @returns void
+             */
+            // eslint-disable-next-line no-unused-vars
+            constructor(year, month, date, hour, minute, second, ms) {
+                // Defensive and verbose to avoid potential harm in passing
+                // explicit undefined when user does not pass argument
+                if (arguments.length === 0) {
+                    super(ClockDate.clock.now);
+                } else {
```
fatso83, line 398:
```suggestion
                // Defensive and verbose to avoid potential harm in passing
                // explicit undefined when user does not pass argument
                if (arguments.length === 0) {
                    super(ClockDate.clock.now);
                } else {
```
Maybe cover this with a test? 🤔

test/fake-timers-test.js
fatso83: This is actually now a stronger comparison. So the test is better. We assert that it is the same function

test/fake-timers-test.js
fatso83: Dug back, found it was just copied over from Sinon ten years ago, so can't be bothered to look deeper:
https://github.com/Oustinger/fake-timers/commit/dbd7d17029f5b258aa14422b74798c844de8bfab#diff-7c9e5e3f3dd46601b73d2d05be43d6d7a33f2f7ccd564494f4eb7b3f4ae915e0R636

The only meaningful bit here about the test is that we check that everything still works as before. So the result of the toString should be the same before and after the removal of the global Date.

test/fake-timers-test.js
fatso83: Just kill the test. It would just be this in the end:

```
node -p 'class FooDate extends Date {}; D=Date; global.Date = function(){};  (new FooDate()) instanceof D'
true
```

fatso83 (changes requested):
Stepan, I took a fresh new look at this and with rested eyes (and a lot of time - sorry), I don't think this looks all that controversial. I get the changes and I think it makes sense. All meaninful tests are running, so I would like to get this in.

If you read this anytime soon, do you think you could be bothered to just implement the _very_ small changes?

fatso83:
@Oustinger This is getting a bit old, but from the comments I see that I basically OK'd it. Any chance you can get it over the finish line? Was fine work.

fatso83 (approved):

fatso83:
As soon as this was out someone found an issue: #504
image

What could be done to get the `instanceof` passing?

benjamingr:
> What could be done to get the instanceof passing? 

We override `Symbol.hasInstance` I guess

benjamingr:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/hasInstance

fatso83:
Ooh, you learn something every day! That's some cool shit. I'll create a fix asap.

fatso83:
Worked like a charm, thanks, Ben!
https://github.com/sinonjs/fake-timers/pull/505

PR #484, opened 2023-10-18
fixes invalid use of arguments in strict mode error

yaruno:

Fix typescript error T1100 which complains about invalid use of term 'arguments' in strict mode.

#### Solution  - optional

Use args instead of arguments in annotation.

Näyttökuva 2023-10-18 133619
Näyttökuva 2023-10-18 133446

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -28,14 +28,14 @@ if (typeof require === "function" && typeof module === "object") {
 /**
  * @callback NextTick
  * @param {VoidVarArgsFunc} callback - the callback to run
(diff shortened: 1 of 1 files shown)


fatso83 (approved):

fatso83:
thanks!

fatso83:
We are not shipping a typescript library and as such you should configure your tsconfig.json to ignore checking this library. So _strictly speaking_, issues in our Jsdoc caught by TSC is not an error, _but_ I can ship a new version regardless, since #479 is not released yet 😄

PR #488, opened 2024-01-03
Fix link to Mochify and adjust description

mantoni:

Mochify became an org with the rewrite and the description was still referring to PhantomJS.


SimenB (approved):
👍

Issue #469, opened 2023-05-22
Mock "timers/promise" module


fatso83:
In #467 @swenzel-arc added the support for the `timers` module in NodeJS. Node also exports a non-callback async/Promise version of that API that needs to be mocked. We can probably very easily piggy-pack on top of the existing work done in the referenced PR. 

Also see the over-arching #418

KholdStare:
Not stale

benjamingr:
This can be closed 🎉

PR #495, opened 2024-08-10
Add support for timers/promises module from nodejs

WhiteAutumn:
#### Purpose (TL;DR)

Fix issue #469 by implementing promise-based wrappers around setTimeout, setImmediate, and setInterval.

#### Solution
This PR takes a similar approach to #467 by @swenzel-arc of detecting the availability of the `timers/promises` module by attempting to import it and then overwriting the functions on the import object, which should be the same object users have, given they do not mess with the module cache. Like the aforementioned PR, the functions are only overwritten when calling `install()` on the `globalObject`.

#### Limitations
* When aborting a timer using an `AbortSignal`, the error thrown is not the same as the one thrown by node. This is due to the `AbortError` error being publicly available. See this [issue](https://github.com/nodejs/node/issues/38361).
* This PR does not touch [scheduler.wait](https://nodejs.org/api/timers.html#timerspromisesschedulerwaitdelay-options) or [scheduler.yield](https://nodejs.org/api/timers.html#timerspromisesscheduleryield) of the `timers/promises` module as they are marked as experimental.

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -1917,25 +1917,21 @@ function withGlobal(_global) {
                             };
 
                             const handle = clock.setTimeout(() => {
-                                if (options.signal) {
-                                    options.signal.removeEventListener(
-                                        "abort",
-                                        abort,
-                                    );
-                                }
+                                options.signal?.removeEventListener(
+                                    "abort",
+                                    abort,
+                                );
 
                                 resolve(value);
                             }, delay);
 
-                            if (options.signal) {
-                                if (options.signal.aborted) {
-                                    abort();
-                                } else {
-                                    options.signal.addEventListener(
-                                        "abort",
-                                        abort,
-                                    );
-                                }
+                            if (options.signal?.aborted) {
+                                abort();
+                            } else {
+                                options.signal?.addEventListener(
+                                    "abort",
+                                    abort,
+                                );
                             }
                         });
                 } else if (nameOfMethodToReplace === "setImmediate") {
@@ -1959,25 +1955,21 @@ function withGlobal(_global) {
                             };
 
                             const handle = clock.setImmediate(() => {
-                                if (options.signal) {
-                                    options.signal.removeEventListener(
-                                        "abort",
-                                        abort,
-                                    );
-                                }
+                                options.signal?.removeEventListener(
+                                    "abort",
+                                    abort,
+                                );
 
                                 resolve(value);
                             });
 
-                            if (options.signal) {
-                                if (options.signal.aborted) {
-                                    abort();
-                                } else {
-                                    options.signal.addEventListener(
-                                        "abort",
-                                        abort,
-                                    );
-                                }
+                            if (options.signal?.aborted) {
+                                abort();
+                            } else {
+                                options.signal?.addEventListener(
+                                    "abort",
+                                    abort,
+                                );
                             }
                         });
                 } else if (nameOfMethodToReplace === "setInterval") {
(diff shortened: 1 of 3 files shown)


src/fake-timers-src.js
```diff
@@ -1872,6 +1891,228 @@ function withGlobal(_global) {
...
+                                reject(options.signal.reason);
+                            };
+
+                            const handle = clock.setTimeout(() => {
+                                if (options.signal) {
+                                    options.signal.removeEventListener(
+                                        "abort",
+                                        abort,
+                                    );
+                                }
```
fatso83, line 1925: Could use optional chaining on most of these options.signal blocks to collapse the indentation.
```suggestion
                                  options.signal.?removeEventListener.?(
                                      "abort",
                                      abort,
                                  );
```

fatso83 (commented):
Very solid work! Love the implementation and the tests. Just had a minor comment. You can see what you mean before I approve. It's just a minor quibble.

src/fake-timers-src.js
WhiteAutumn: Sure! I'm all for optional chaining. I'll go through and replace instances of `if (options.signal)` with optional chaining.

benjamingr (approved):
There are some leaks in the signal (e.g. we don't remove the event listener when fake timers are uninstalled) but generally that's fine for the first version so lgtm.

fatso83:
Good note about the removal of event listeners. When I do any development on Sinon projects, I tend to make use of the watch feature of the test runner. That's where stuff like not cleaning up properly tends to show itself.

If it's not too much problem, I'd like to see that being done.

WhiteAutumn:
Sure, I can take a look at it. Just so we're on the same page, we're talking about the scenario where while a timer is running, an abort signal has been passed, and the clock is uninstalled before the timer fires, leading to it never firing and the therefore the abort signal listener never being removed?

This is something I have missed, and the listeners should absolutely be removed in that scenario, but for what it's worth I don't think it would lead to memory leaks while running tests in watch mode. As long as no references remain to the abort signal, it should be garbage collected and the event listener along with it.

As to how to solve this, I would like a second opinion on how to approach it. The way I see it this would require something akin to an `onUninstall` event that the promise based implementations can listen to and then remove the abort event listeners when fired. It doesn't feel great adding more event listeners to solve hanging event listeners but I can't think of any other way of doing it. What do you think?

fatso83:
You know what, I don't have available mental capacity (or time) at the moment to dig deeply into this, and if it really isn't an issue in practice, then I have no issues merging this now. There's always room for another PR later, if you find a nice way that isn't too complicated.

fatso83 (approved):

fatso83:
Thanks for the reminder :)

benjamingr:
> This is something I have missed, and the listeners should absolutely be removed in that scenario, but for what it's worth I don't think it would lead to memory leaks while running tests in watch mode. As long as no references remain to the abort signal, it should be garbage collected and the event listener along with it.

References can remain to the signal (e.g. the signal is the result of `AbortSignal.any` of a local signal and a process-level signal.

It's fine for it to happen at a follow up PR. Generally you did cleanup a lot better than I expected in this PR in the first place and it's a relatively small miss. I review a lot of AbortSignal library code so this is a compliment :)

> As to how to solve this, I would like a second opinion on how to approach it. The way I see it this would require something akin to an onUninstall event that the promise based implementations can listen to and then remove the abort event listeners when fired. It doesn't feel great adding more event listeners to solve hanging event listeners but I can't think of any other way of doing it. What do you think?

We keep a list/map or all signal and event listeners associated with the clock instance and when the clock is uninstalled we call that?

```js
// Somewhere when an event listener is added
signal.addEventListener('abort', listener, options); 
clock.listners.push({ signal, listener });

// In uninstall
clock.listeners.forEach(({ signal, listener }) => signal.removeEventListener(listener))
```

WhiteAutumn:
> References can remain to the signal (e.g. the signal is the result of AbortSignal.any of a local signal and a process-level signal.
Ohh, I see. Yeah then it's definitely something that would be good to get fixed before a release.

> We keep a list/map or all signal and event listeners associated with the clock instance and when the clock is uninstalled we call that?
Thanks for the input. I'll proceed with that kind of a solution and make a separate PR.

fatso83:
@u873838 Thanks to White Autumns swift second PR this is already out.

PR #497, opened 2024-08-23
Fix listener leak for timers/promises

WhiteAutumn:
#### Purpose (TL;DR)
Fix abort event listener leak discussed in #495 by adding `abortListenerMap` to keep track of active abort listeners.

#### Solution
The leak is fixed by adding a map with listener functions as keys and the abort signals as values. During uninstall it iterates over the map and removes the listeners. If a timer is aborted or finishes on its own the listener signal pair will be removed from the map.

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -99,6 +99,7 @@ if (typeof require === "function" && typeof module === "object") {
  * @property {boolean} [shouldClearNativeTimers] inherited from config
  * @property {{methodName:string, original:any}[] | undefined} timersModuleMethods
  * @property {{methodName:string, original:any}[] | undefined} timersPromisesModuleMethods
+ * @property {Map<function(): void, AbortSignal>} abortListenerMap
  */
 /* eslint-enable jsdoc/require-property-description */
 
@@ -967,6 +968,11 @@ function withGlobal(_global) {
         // Prevent multiple executions which will completely remove these props
         clock.methods = [];
 
+        for (const [listener, signal] of clock.abortListenerMap.entries()) {
+            signal.removeEventListener("abort", listener);
+            clock.abortListenerMap.delete(listener);
+        }
+
         // return pending timers, to enable checking what timers remained on uninstall
         if (!clock.timers) {
             return [];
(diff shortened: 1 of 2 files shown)


src/fake-timers-src.js
```diff
@@ -1789,6 +1795,8 @@ function withGlobal(_global) {
             return uninstall(clock, config);
         };
 
+        clock.abortListenerMap = new Map();
```
WhiteAutumn, line 1798: Made this always initialize instead of putting it behind the check of `timers/promises` being available as it is conceivable that in the future other parts of the code may also want to unregister abort signal event listeners on uninstall

src/fake-timers-src.js
```diff
@@ -967,6 +968,11 @@ function withGlobal(_global) {
         // Prevent multiple executions which will completely remove these props
         clock.methods = [];
 
+        for (const [listener, signal] of clock.abortListenerMap.entries()) {
+            signal.removeEventListener("abort", listener);
+            clock.abortListenerMap.delete(listener);
```
WhiteAutumn, line 973: Modifying while iterating over maps appears to be [safe](https://stackoverflow.com/questions/35940216/es6-is-it-dangerous-to-delete-elements-from-set-map-during-set-map-iteration)

src/fake-timers-src.js
```diff
@@ -1903,21 +1913,30 @@ function withGlobal(_global) {
                             };
 
                             const handle = clock.setTimeout(() => {
-                                options.signal?.removeEventListener(
-                                    "abort",
-                                    abort,
-                                );
+                                if (options.signal) {
```
WhiteAutumn, line 1916: Opted to move back to regular if checks as opposed to optional chaining for consistency and readability, as you can't optional-chain adding / removing signals to `abortListenerMap`

benjamingr (approved):

src/fake-timers-src.js
benjamingr: It is

fatso83:
Great addition. Both this and the previous PR is very nice work!

PR #501, opened 2024-08-26
Convert to ESM

### LINKED ISSUE #489: Skypack bundle cannot be loaded in the browser

-   FakeTimers version : 11.2.2
-   Environment : macOS 14.2.1, Firefox 121.0.1
-   Example URL : https://output.jsbin.com/xijaquyico
-   Other libraries you are using: none

FakeTimers module is loaded.

**What actually happens**
An error is logged and the module does not load:
```
Uncaught Error: [Package Error] "timers" does not exist. (Imported by "@sinonjs/fake-timers").
    <anonymous> https://cdn.skypack.dev/error/node:timers?from=@sinonjs/fake-timers:14
```

**How to reproduce**

Go to https://output.jsbin.com/xijaquyico and open the JavaScript console to see the error. The module is loaded like this:

```html
  <script type="module">
    import FakeTimers from "https://cdn.skypack.dev/@sinonjs/fake-timers@v11.2.2";
  </script>
```

fatso83:

Fix #489 (Skypack import) by converting to ESM.

This change should not be handled without forethought ... To be able to do this, we probably need
to convert the main `sinon` package to ESM as well.

### DIFF
--- a/integration-test/fake-clock-integration-test.js
+++ b/integration-test/fake-clock-integration-test.js
@@ -1,25 +1,7 @@
-"use strict";
-
-let jsdom;
-
-if (typeof require === "function" && typeof module === "object") {
-    try {
-        jsdom = require("jsdom");
-    } catch (e) {
-        // ignored
-    }
-}
-
-if (!jsdom) {
-    // eslint-disable-next-line no-console
-    console.warn("JSDOM is not supported in the current environment.");
-
-    return;
-}
-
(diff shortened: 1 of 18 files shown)


fatso83:
ESM would never have fixed the Skypack issue, but 3dbfaa83a4c54c3103c5265597bbdc6b8b4b4e15 did. We can still do the ESM conversion, though, now that that issue is out of the way :)

PR #509, opened 2024-09-26
feat: Add additional auto advance time controls

atscott:
#### Purpose (TL;DR)

This adds a new mode for automatically advancing time that moves more quickly than the existing shouldAdvanceTime, which uses real time. It also adds the ability to modify the initial values that were used for shouldAdvanceTime and advanceTimeDelta. `shouldAdvanceTime` is useful in some situations, but is not useful for solving the problem that testers usually install mock clocks for (to speed up time).

#### Background 

Testing with mock clocks can often turn into a real struggle when dealing with situations where some work in the test is truly async and other work is captured by the mock clock.

In addition, when using mock clocks, testers are always forced to write tests with intimate knowledge of when the mock clock needs to be ticked. Oftentimes, the purpose of using a mock clock is to speed up the execution time of the test when there are timeouts involved. It is not often a goal to test the exact timeout values. This can cause tests to be riddled with manual advancements of fake time. It ideal for test code to be written in a way that is independent of whether a mock clock is installed or which mock clock library is used. For example:

```
document.getElementById('submit');
// https://testing-library.com/docs/dom-testing-library/api-async/#waitfor
await waitFor(() => expect(mockAPI).toHaveBeenCalledTimes(1))
```

When mock clocks are involved, the above may not be possible if there is some delay involved between the click and the request to the API. Instead, developers would need to manually tick the clock beyond the delay to trigger the API call.

This is different from the existing `shouldAdvanceTime` in the following ways:

`shouldAdvanceTime` is essentially `setInterval(() => clock.tick(ms), ms)` while this feature is `const loop = () => setTimeout(() => clock.nextAsync().then(() => loop()), 0);`

There are two key differences between these two:

1. `shouldAdvanceTime` uses `clock.tick(ms)` so it synchronously runs all timers inside the "ms" of the clock queue. This doesn't allow the microtask queue to empty between the macrotask timers in the clock whereas something like `tickAsync(ms)` (or a loop around `nextAsync`) would. This could arguably be considered a fixable bug in its implementation
2. `shouldAdvanceTime` uses real time to advance the same amount of real time in the mock clock. The way I understand it, this feels somewhat like "real time with the opportunity to advance more quickly by manually advancing time". This would be quite different: It advances time as quickly possible and as far as necessary. Without manual ticks, `shouldAdvanceTime` would only be capabale of automatically advancing as far as the timeout of the test and take the whole real time of the test timeout. In contrast, `setTickMode({mode: "nextAsync"})` can theoretically advance infinitely far, limited only by processing speed. Somewhat similar to the [--virtual-time-budget](https://developer.chrome.com/docs/chromium/headless#--virtual-time-budget) feature of headless chrome.

In addition to the "quick mode" of `shouldAdvanceTime`, this also adds the ability to modify the initially configured values for shouldAdvanceTime and advanceTimeDelta.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,12 @@ setTimeout(() => {
 }, 50);
 ```
 
+In addition to the above, mocked time can be configured to advance more quickly
+using `clock.setTickMode({ mode: "nextAsync" });`. With this mode, the clock
+advances to the first scheduled timer and fires it, in a loop. Between each timer,
+it will also break the event loop, allowing any scheduled promise
+callbacks to execute _before_ running the next one.
+
 ## API Reference
 
 ### `var clock = FakeTimers.createClock([now[, loopLimit]])`
@@ -180,12 +186,24 @@ The following configuration options are available
 
 Allows configuring how the clock advances time, automatically or manually.
 
-There are 2 different types of modes for advancing timers:
+There are 3 different types of modes for advancing timers:
 
 -   `{mode: 'manual'}`: Timers do not advance without explicit, manual calls to the tick
     APIs (`jest.advanceTimersToNextTimer`, `jest.runAllTimers`, etc). This mode is equivalent to `false`.
+-   `{mode: 'nextAsync'}`: The clock will continuously break the event loop, then run the next timer until the mode changes.
+    As a result, tests can be written in a way that is independent from whether fake timers are installed.
+    Tests can always be written to wait for timers to resolve, even when using fake timers.
 -   `{mode: 'interval', delta?: <number>}`: This is the same as specifying `shouldAdvanceTime: true` with an `advanceTimeDelta`. If the delta is
     not specified, 20 will be used by default.
+
+The 'nextAsync' mode differs from `interval` in two key ways:
+
+1.  The microtask queue is allowed to empty between each timer execution,
+    as would be the case without fake timers installed.
+1.  It advances as quickly and as far as necessary. If the next timer in
+    the queue is at 1000ms, it will advance 1000ms immediately whereas interval,
+    without manually advancing time in the test, would take `1000 / advanceTimeDelta`
+    real time to reach and execute the timer.
 
 ### `var id = clock.setTimeout(callback, timeout)`
 
--- a/README.md
+++ b/README.md
@@ -176,6 +176,17 @@ The following configuration options are available
 | `config.shouldClearNativeTimers` | Boolean     | false                                                                                                                                                                                                                          | tells FakeTimers to clear 'native' (i.e. not fake) timers by delegating to their respective handlers. These are not cleared by default, leading to potentially unexpected behavior if timers existed prior to installing FakeTimers.                                   |
 | `config.ignoreMissingTimers`     | Boolean     | false                                                                                                                                                                                                                          | tells FakeTimers to ignore missing timers that might not exist in the given environment                                                                                                                                                                                |
 
+### `clock.setTickMode(mode)`
+
+Allows configuring how the clock advances time, automatically or manually.
+
+There are 2 different types of modes for advancing timers:
+
+-   `{mode: 'manual'}`: Timers do not advance without explicit, manual calls to the tick
+    APIs (`jest.advanceTimersToNextTimer`, `jest.runAllTimers`, etc). This mode is equivalent to `false`.
+-   `{mode: 'interval', delta?: <number>}`: This is the same as specifying `shouldAdvanceTime: true` with an `advanceTimeDelta`. If the delta is
+    not specified, 20 will be used by default.
+
 ### `var id = clock.setTimeout(callback, timeout)`
 
 Schedules the callback to be fired once `timeout` milliseconds have ticked by.
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -16,6 +16,25 @@ if (typeof require === "function" && typeof module === "object") {
 }
 
 /**
+ * @typedef {"manual" | "interval"} TickMode
+ */
+
+/**
+ * @typedef {object} ManualTickMode
+ * @property {"manual"} mode
+ */
+
+/**
+ * @typedef {object} IntervalTickMode
+ * @property {"interval"} mode
+ * @property {number} [delta]
+ */
+
+/**
+ * @typedef {IntervalTickMode | ManualTickMode} TimerTickMode
+ */
+
+/**
  * @typedef {object} IdleDeadline
  * @property {boolean} didTimeout - whether or not the callback was called before reaching the optional timeout
  * @property {function():number} timeRemaining - a floating-point value providing an estimate of the number of milliseconds remaining in the current idle period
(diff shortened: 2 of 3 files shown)


fatso83:
Thanks for this, I will have to check out the code at a later time. 

> testers are always forced to write tests with intimate knowledge of when the mock clock needs to be ticke

just to check ... You are aware of `await clock.runAllAsync()` and `await clock.runToLastAsync()`, right? They often solve the issues of needing intimate knowledge about when timers are being fired and when Promises are needing to be resolved.

atscott:
> just to check ... You are aware of await clock.runAllAsync() and await clock.runToLastAsync(), right? They often solve the issues of needing intimate knowledge about when timers are being fired and when Promises are needing to be resolved.

Yes, these are quite useful! Here are some of my thoughts where these still cause some test friction that this feature would help resolve:

#### When mock clocks are installed, it's not "safe" to `await` just anything

When writing tests, you have to know that whatever you're `awaiting` will not wait for any timers to complete. Either that, or you need to make sure that you flush timers _before_ awaiting any promises. It's especially difficult if the function _causes_ a timer to be queued and then waits for it. The way around this would be to save the promise instead of `await`ing it, then flush timers, then await the promise (pretty awkward).

#### When mock clocks are installed, test utilities need to know about it

Today, test utilities that perform async actions need to be written to account for mock clocks. In the above situation, if the function we're talking about was some test harness or test utility, we could tell it how to tick the clock so it can advance time internally rather than us having to do it in the test body.

#### Concrete example: Testing Gmail's "undo" send button
Consider testing the "undo" send button feature in Gmail. Conceptually, what you want to do in the test is "click send, wait for the toast to appear, then click the 'undo' button". The undo button doesn't appear synchronously and would require running some timers to make it appear. Using `runAllAsync` would advance time to the point where the email is fully sent and the toast disappears. You'd have to write the test to tick some amount of time for it to appear, but not long enough for it to disappear.

```
await sendButton.click();
await clock.tickAsync(100);
const undoButton = await page.undoButton();
await undoButton.click();
// follow-up expectations
```

With the auto-clicking mock clock, this test could instead look much more like it would in an e2e environment (and in fact, could even use the same test body, with the page harnesses swapped out underneath which uses WebDriver or the framework testing library depending on environment):

```
await sendButton.click();
// Maybe undoButton getter polls for the button with a timeout
const undoButton = await page.undoButton();
await undoButton.click();
```

----

In each of the above situations, an automatically advancing mock clock would make testing much more reasonable. It's certainly not always the answer and it may also still be useful to manually flush timers (i.e. in the "undo" button example, flushing timers after clicking undo and then asserting that the RPC to send the message never happened). While `runAllAsync` and `runToLastAsync` are absolutely useful, they still require the test author to actually call them to advance time. It's less intimate knowledge of the timer, but it is still needing to know that the promise doesn't resolve in the event loop.

atscott:
Hi @fatso83, is there anything more you need from me here? 

We feel like this is quite a powerful feature, unlocking all the benefits of a fast mock clock without needing to change anything else about how a test is written. One might discover a slow test suite and only need to install an auto-ticking mock clock to speed things up.

fatso83:
Sorry about that. Was hoping someone else on the maintainer team would pick it up, and I have simply forgotten about it 😄 I will have a new look tomorrow (and remind me if there's no update by next week), as I no longer remember the details ...

atscott:
@fatso83 I updated the commit to split the new "setTickMode" feature out from the new "nextAsync" mode. Hopefully this makes it easier to review.

atscott:
@fatso83 Is there any existing/accepted way in the Sinon fake timers code to check if it's running in node? `setImmediate` is a pretty fairly significant improvement over `setTimeout`. I found that it's often several times faster than `setTimeout(0)` and made that change to the jasmine version of this PR (https://github.com/jasmine/jasmine/pull/2058)

fatso83:
@atscott Hi, Scott. Seeing this hanging for such a long time is embarrassing, but I have almost not been touching JS for the last six months, so it's been hard justifying working on this. Hopefully, you have been able to fork this for your own needs in the mean time or patch the code to not be blocked. I'll try to give it some love soon.

I am on my phone now, so giving great feedback is awkward, but we already do feature detection in the first few lines. That means you don't need to know if you are on node to use setImmediate, you can just use the existing check to see if it's present and just setImmediate directly.

benjamingr (commented):
The actual code is fine but I don't think it's a particularly good feature since it breaks order and correctness. It's also a big patch.

I'm not blocking.

atscott:
> Seeing this hanging for such a long time is embarrassing, but I have almost not been touching JS for the last six months, so it's been hard justifying working on this. Hopefully, you have been able to fork this for your own needs in the mean time or patch the code to not be blocked. I'll try to give it some love soon.

No worries, I totally get it. 

> you can just use the existing check to see if it's present and just setImmediate directly

I had some concern with browsers that happened to implement it but I suppose it would probably still just work as intended if they do. Plus, those very few that have it are likely far out of support. 

> I don't think it's a particularly good feature since it breaks order and correctness

Would you mind expanding on this a bit more? Do you have an example where this breaks down in a place that it doesn’t already with existing APIs?

> It's also a big patch

That’s fair. Quite a bit of the change is in the first commit that only adds the ability to update the ticking mode between the two options that already exist.

The other large portion comes from the need for the MessageChannel workaround for timeout throttling. Aside from that, it’s mostly white space changes around the async tick methods. Have you tried turning off the white space in the diff settings?

If you prefer, I could likely reduce the raw diff by removing the ability to change modes after the clock was created and also remove some longer explanatory comments.

fatso83:
I have forgotten about this, so this was a great reminder. I stopped for the same reasons that Andrew Asked about: I was uncertain about the order and correctness. 

@Uzlopak , can you expand on your comment: "it was unreliable and made the tests randomly hang."?

I assume this meant you have checkout out this fork and used it to test out the new API. Yes?

Can you please post some example code that uses the new time controls that will show this randomly failing behavior? Without this, it is _very_ unclear if this is user error (wrong use of `runAllAsync`) or an inherent design issue.

Uzlopak:
I did not use your code. I used runAllAsync..

https://github.com/nodejs/undici/pull/4526/files#diff-bf38571af1c6adc939fb9ea1e5af94cd46ae0933dd9298aaa6cef7b5835fd2a3

Replacing 

```js
  const timer = setTimeout(() => {
    clock?.tick(100)
    timer.refresh()
  }, 0).unref()

  const clock = FakeTimers.install()
  after(() => clock.uninstall())
```

with

```js
  const clock = FakeTimers.install()
  clock.runAllAsync()
```

and then running the test file multiple times, than one of the runs will get stuck.

atscott:
> Sorry for keeping it open for such a long time - hope you see why.

No worries at all! As a maintainer of an open source library myself, I appreciate the thoughtfulness and caution. I could have easily understood waiting another year, even if only to see how things play out in the jasmine version (`jasmine.clock().autoTick()`) that was released in 5.7.

Issue #510, opened 2024-10-17
ClockTime constructor name is different, breaking mongoose Date checks


alexanderweiss:
> We understand you have a problem and are in a hurry, but please provide us with some info to make it much more likely for your issue to be understood, worked on and resolved quickly.

-   FakeTimers version : 13.0.2
-   Environment : node 20.18 (macOS)
-   Other libraries you are using: sinon 19.0.2, mongoose 8.7.1

**What did you expect to happen?**
We're using Sinon together with mongoose. When using mongoose's `toJSON` and `toObject` to turn mongoose documents into plain objects while faking timers I'd expect fields with Date objects to remain Date objects (or at least look like them). This is the case in fake-timers 11.2.2.

**What actually happens**
While faking timers, when calling `toJSON` or `toObject` on a mongoose document, it turns any Date fields into a number (unix timestamp). This seems to happen because [mongoose determines whether an object is a Date or not using the constructor name](https://github.com/Automattic/mongoose/blob/0752c5daea10df87183a27ed7a1e2c3be33ef52a/lib/helpers/clone.js#L77-L89).

**How to reproduce**
```js
import mongoose from 'mongoose';
import sinon from 'sinon';

const DocModel = mongoose.model('doc',  new mongoose.Schema({ someDate: { type: Date } }));

const doc = new DocModel({ someDate: new Date() })

const plainDocBefore = doc.toObject()
console.log(plainDocBefore.someDate) // plainDocBefore.someDate is a Date object

sinon.useFakeTimers({ now: new Date(), toFake: ['Date', 'hrtime'] });

const plainDocAfter = doc.toObject()
console.log(plainDocAfter.someDate) // plainDocAfter.someDate is a unix timestamp (number)
```

Hope this is the right place to report this.

fatso83:
Hi, great report. We are trying to do semantic releases, which means that when going from version 11 to 12, you will know there are breaking changes. So there are at least two breaking changes between version 11 and 13, of which [this is the most important change in version 12](https://github.com/sinonjs/fake-timers/blob/main/CHANGELOG.md#1200--2024-08-24).

So currently it works as intended, but that does not mean we cannot do stuff to make life easier for people, if it is not a big deal. We did this in #505, for instance, to make `instanceof Date` checks pass.

[...]

PR #512, opened 2024-10-21
Fix constructor-based checks for fake Date no longer pass after installing

alexanderweiss:
Fixes #511 (fix for #510) no longer working after the clock had been installed and global `Date` assigned to the fake.

Once `ClockDateProxy` was installed to global `Date` (to my surprise) the constructor reference on it was 'updated' to point to `ClockDateProxy`.

Thanks again for the help!

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -450,7 +450,7 @@ function withGlobal(_global) {
 
                 // ensures identity checks using the constructor prop still works
                 // this should have no other functional effect
-                this.constructor = Date;
+                this.constructor = NativeDate;
             }
 
             static [Symbol.hasInstance](instance) {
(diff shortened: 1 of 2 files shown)


src/fake-timers-src.js
```diff
@@ -450,7 +450,7 @@ function withGlobal(_global) {
 
                 // ensures identity checks using the constructor prop still works
                 // this should have no other functional effect
-                this.constructor = Date;
+                this.constructor = NativeDate;
```
fatso83, line 453: Good catch, this was naively working in the tests due to it being the same when not overwritten. This is what I _meant_ to have written 😄

fatso83 (approved):

src/fake-timers-src.js
alexanderweiss: 👍 I was initially very surprised, but makes sense now — since this happens every time a fake Date is created.

fatso83: patch version out shortly

PR #513, opened 2024-10-28
Ensure checks based Object.keys length pass for fake Dates

alexanderweiss:
Fixes sinon.match / samsam equality checks failing between `Date` and fake dates.

#### Background
Ran into another small issue with the constructor: samsam's `deepEqual` [checks the number](https://github.com/sinonjs/samsam/blob/1b141a7cb9d54b0d8a1f20812c2db4bc1f1e53a3/lib/deep-equal.js#L154) of properties on the object, and the constructor [should not be enumerable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/constructor) so there should be 0 keys, but fake Dates had one now.

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -450,7 +450,10 @@ function withGlobal(_global) {
 
                 // ensures identity checks using the constructor prop still works
                 // this should have no other functional effect
-                this.constructor = NativeDate;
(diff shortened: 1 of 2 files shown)


alexanderweiss:
Thanks! Our tests were all green now (When adding an override for the latest fake-timers version)

PR #515, opened 2024-12-03
fix: return fake version for `performance.timeOrigin`

### LINKED ISSUE #514: Support `performance.timeOrigin` property

-   FakeTimers version : 13.0.5
-   Environment : MacOS
-   Other libraries you are using: jest, @sentry/react

I expected that `performance.timeOrigin` (https://developer.mozilla.org/en-US/docs/Web/API/Performance/timeOrigin) would return a valid value with fake timers enabled in jest.

**What actually happens**

`performance.timeOrigin` is set to NOOP [here](https://github.com/sinonjs/fake-timers/blob/7861c930e8f9aaa9d227431623cc82347aff0291/src/fake-timers-src.js#L1843).

**How to reproduce**

```bash
git clone https://github.com/rriski/fake-timers-timeOrigin-repro.git
npm run test
# --> test fails
# Uncomment `doNotFake` in jest.config.js
npm run test
# --> test passes
```

**Additional information**

* Related to #493
* `@sentry/react` uses the performance API to set the session starting time, see [session.ts](https://github.com/getsentry/sentry-javascript/blob/6331d7ced7373e3f351ffacb71af54392c19acda/packages/core/src/session.ts#L14)

rriski:
Fixes #514

Return valid value for `performance.timeOrigin` instead of `NOOP`. This fixes an issue in tests that depend on `performance.timeOrigin` having a valid value.

See #514

#### Solution  - optional

Set `performance.timeOrigin` to a dummy value. I'm not sure if 0 is better than setting the value to `clock.now`. When testing `performance.timeOrigin` in Chrome Console I can see that the value is set to the timestamp when the window was opened:

```
> globalThis.performance.timeOrigin
1733227796841
```

There is further information in MDS docs on how the value is supposed to be set: https://developer.mozilla.org/en-US/docs/Web/API/Performance/timeOrigin#value

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -1849,9 +1849,7 @@ function withGlobal(_global) {
                 clock.performance.measure = (name) =>
                     new FakePerformanceEntry(name, "measure", 0, 100);
                 // ensure `timeOrigin` returns a valid value
-                clock.performance.timeOrigin = config.now
-                    ? getEpoch(config.now)
-                    : new Date().getTime();
+                clock.performance.timeOrigin = getEpoch(config.now);
             } else if ((config.toFake || []).includes("performance")) {
                 return handleMissingTimer("performance");
             }
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -23,7 +23,6 @@ if (typeof require === "function" && typeof module === "object") {
 
 /**
  * Queues a function to be called during a browser's idle periods
- *
  * @callback RequestIdleCallback
  * @param {function(IdleDeadline)} callback
  * @param {{timeout: number}} options - an options object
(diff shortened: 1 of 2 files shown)


src/fake-timers-src.js
```diff
@@ -1848,6 +1848,8 @@ function withGlobal(_global) {
                     new FakePerformanceEntry(name, "mark", 0, 0);
                 clock.performance.measure = (name) =>
                     new FakePerformanceEntry(name, "measure", 0, 100);
+                // ensure `timeOrigin` returns a valid value
+                clock.performance.timeOrigin = 0;
```
fatso83, line 1852: > If the current [Document](https://developer.mozilla.org/en-US/docs/Web/API/Document) is the first one loaded in the Window, the time origin is the time at which the browser context was created.

Best fit is likely the installed time

rriski: Would this work?

```js
clock.performance.timeOrigin = config.now ? getEpoch(config.now) : new Date().getTime();
```

fatso83: A bit redundant, given that `getEpoch` already does just that. I trimmed it away and also made the test a bit tighter/more specific.

rriski: Thank you!

PR #527, opened 2025-09-18
docs: Update `setTickMode` docs to mention sinon APIs, not jest

atscott:
I had copied my implementation from jest into sinon. Sorry for missing this.


fatso83 (approved):
I actually noticed it, but thought I saw it listed alongside the Sinon ones, which I thought was OK, given that Jest uses this library. Did not realise it was _only_ Jest.

PR #549, opened 2026-03-31
fix: memory leak and performance issues from #532

fatso83:

See #532 for the detailed analysis. This PR attempts to solve all findings, and includes a massive speedup, showcasing going from 15 seconds to 7 ms for a given case of 10k timers.

## Summary of Fixes:

1. BUG-1: Memory Leak via Timer Refresh: Refactored the timer handle object (the res object returned when addTimerReturnsObject is true) to use this instead of closing over itself. This breaks the circular reference that prevented garbage collection when users held references to timer handles.
2. BUG-2: O(N) Performance Degradation: Implemented a Min-Heap (TimerHeap) to store and retrieve timers. This replaces the previous O(N) linear search for the next timer with O(1) peek and O(log N) insertion/removal, significantly improving performance for test suites with thousands of timers.
3. BUG-3: Logic Error in Uninstall: Corrected the typo and logic in uninstall and hijackMethod. Instead of relying on a potentially fragile property on the fake function, the cleanup logic now more robustly verifies and restores global properties.
4. BUG-4: Integer Overflow Risk: Implemented a bounded uniqueTimerId that wraps around at Number.MAX_SAFE_INTEGER and includes a collision check to ensure active timers are never overwritten.
5. Special Case - requestIdleCallback: Fixed a regression found during testing where a single requestIdleCallback timer should fire even if outside the current tick range if it's the only thing scheduled.

## Why it's more performant:

In the original implementation, fake-timers used an O(N) linear search for every timer operation. To find the next timer to fire, it would
iterate over every active timer ID:

```javascript
// ORIGINAL (Simplified) - O(N) linear search
function firstTimer(clock) {
    for (id in timers) { // Iterates over every single timer
        if (!timer || compareTimers(timer, timers[id]) === 1) {
            timer = timers[id];
        }
    }
    return timer;
}
```

In the fixed implementation, I introduced a Min-Heap (TimerHeap). This allows the clock to:
 - Find the next timer in O(1) time (always at the top of the heap).
 - Maintain the heap structure in O(log N) time for insertions and removals.

## Why it's more robust:
 1. Memory: By refactoring the res object to use this instead of closing over itself, we eliminated the circular reference that caused memory leaks when timer handles were held.
 2. ID Safety: The new uniqueTimerId logic wraps safely at Number.MAX_SAFE_INTEGER and includes a collision check, preventing timer ID conflicts that could occur in long-running processes.
 3. Correctness: Fixed the typo hadOwnProperty which could lead to incorrect global property restoration during uninstall().

## How to verify manually:
You can run the comparison yourself using the following script. The difference is night and day: for 10,000 timers, the original code takes 15 seconds to finish, while the optimized version takes only 7 milliseconds.

## Performance test verification
```javascript
git checkout main

cat > perf-test.js << EOF
"use strict";

const fakeTimers = require("./src/fake-timers-src.js");
const clock = fakeTimers.createClock();

const NUM_TIMERS = 10000;

console.log(`--- Performance Test with ${NUM_TIMERS} timers (Original) ---`);

const startCreate = Date.now();
for (let i = 0; i < NUM_TIMERS; i++) {
    clock.setTimeout(() => {}, i);
}
const endCreate = Date.now();
console.log(`Time to create ${NUM_TIMERS} timers: ${endCreate - startCreate}ms`);

const startTick = Date.now();
clock.tick(NUM_TIMERS);
const endTick = Date.now();
console.log(`Time to tick(${NUM_TIMERS}): ${endTick - startTick}ms`);

// Test scenario where we have many timers but only tick a small amount
const clock2 = fakeTimers.createClock();
for (let i = 0; i < NUM_TIMERS; i++) {
    clock2.setTimeout(() => {}, NUM_TIMERS * 2);
}

console.log(`\n--- Performance Test: tick(1) x 100 with ${NUM_TIMERS} pending timers ---`);
const startTickSmall = Date.now();
for (let i = 0; i < 100; i++) {
    clock2.tick(1);
}
const endTickSmall = Date.now();
console.log(`Time for 100 ticks: ${endTickSmall - startTickSmall}ms`);
EOF

# this takes on the order of 15 seconds
node perf-test.js

git checkout fix-issue-532-new 

# this should take about 7 milliseconds
node perf-test.js
```

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -21,18 +21,18 @@ if (typeof require === "function" && typeof module === "object") {
 
 /**
  * @typedef {object} NextAsyncTickMode
- * @property {"nextAsync"} mode
+ * @property {"nextAsync"} mode - runs timers one macrotask at a time
  */
 
 /**
  * @typedef {object} ManualTickMode
- * @property {"manual"} mode
+ * @property {"manual"} mode - advances only when the caller explicitly ticks
  */
 
 /**
  * @typedef {object} IntervalTickMode
- * @property {"interval"} mode
- * @property {number} [delta]
+ * @property {"interval"} mode - advances automatically on a native interval
+ * @property {number} [delta] - interval duration in milliseconds
  */
 
 /**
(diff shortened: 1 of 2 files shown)


src/fake-timers-src.js
```diff
@@ -996,7 +1100,7 @@ function withGlobal(_global) {
                     _global[method] = clock[`_${method}`];
                 }
             } else {
-                if (_global[method] && _global[method].hadOwnProperty) {
+                if (clock[method] && clock[method].hasOwnProperty) {
```
BridgeAR, line 1103: This seems to be a mistake? I believe it should still be

```suggestion
                if (clock[method] && clock[method].hadOwnProperty) {
```

The same where it was assigned.

src/fake-timers-src.js
```diff
@@ -1049,7 +1151,7 @@ function withGlobal(_global) {
      * @param {Clock} clock
      */
     function hijackMethod(target, method, clock) {
-        clock[method].hadOwnProperty = Object.prototype.hasOwnProperty.call(
+        clock[method].hasOwnProperty = Object.prototype.hasOwnProperty.call(
```
BridgeAR, line 1154: This seems to be a regression, where the result is now assigning a boolean.

```suggestion
        clock[method].hadOwnProperty = Object.prototype.hasOwnProperty.call(
```

BridgeAR:
@fatso83 PTAL at my comments. This change seems to regress a few things.

This has come up in our tests after sinon incorporating the new version.

fatso83:
@BridgeAR Hi, can you be more specific in what regressions you are talking about? This is just internals, so I am not sure where the _regression_ is. Is there something not working that used to work? If so, please file a bug! Otherwise, I am not sure what you are implying 😄 

BTW I am not saying there is a regression, just that I do not understand how it manifests 😸 All tests pass, which is our measure.

BridgeAR:
@fatso83 I just opened https://github.com/sinonjs/fake-timers/pull/576 for it. I hope that explains the bug (and fixes it) :)

PR #561, opened 2026-04-10
Restore Date object support broken by PR #556

adspacheco:
                                                                                                                                            
Fix cross-realm Date regression introduced in [#556](https://github.com/sinonjs/fake-timers/pull/556).                                                                                       
                                                                                                                                            
The public API accepts `Date | number`, but the change in #556 from `typeof epoch.getTime === "function"` to `instanceof Date` broke support for Date objects from different realms (iframes, vm contexts, etc.), since `instanceof` fails across realms.                      
                                                                                                                                            
Error: `TypeError: now should be milliseconds since UNIX epoch`                                                                           
                                                                                                                                            
Reproduction:                                                                                                                             
                                                                                                                                            
```js                                                                                                                                     
const vm = require("vm");                                                                                                                 
const FakeTimers = require("@sinonjs/fake-timers");                                                                                       
const crossRealmDate = vm.runInNewContext("new Date('2026-04-02')");                                                                      
const clock = FakeTimers.install();                                                                                                       
clock.setSystemTime(crossRealmDate); // throws TypeError                                                                                  
```      
                                                                                                                                     
#### Solution                                                                                                                                  
                                                                                                                                            
Restore the original duck-typing approach `(typeof epoch.getTime === "function")` that worked before #556.

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -639,7 +639,7 @@ function withGlobal(_global) {
         if (!epoch) {
             return 0;
         }
-        if (epoch instanceof Date) {
+        if (typeof epoch.getTime === "function") {
             return epoch.getTime();
         }
         if (typeof epoch === "number") {


fatso83:
added the missing regression test in aa15fde41f5bda6e118adcffdcbd012d52ec4bcc for you, as I introduced the regression 😉

fatso83:
Unfortunately, this fix was incomplete, as the typing now breaks. This is what I get for not ensuring the types are built as part of the github actions. At least I know that now 😄 
```
src/fake-timers-src.js:642:26 - error TS2339: Property 'getTime' does not exist on type 'number | Date'.
  Property 'getTime' does not exist on type 'number'.

642         if (typeof epoch.getTime === "function") {
                             ~~~~~~~

src/fake-timers-src.js:643:26 - error TS2339: Property 'getTime' does not exist on type 'number | Date'.
  Property 'getTime' does not exist on type 'number'.

643             return epoch.getTime();
                             ~~~~~~~


Found 2 errors in the same file, starting at: src/fake-timers-src.js:642

```

I'll get it fixed.

fatso83:
types are now always built and the type regression fixed in d43b719

release 15.3.1 out in a few jiffies

PR #569, opened 2026-05-05
feat: add support for `Temporal` API

### LINKED ISSUE #335: Support `temporal` TC39 proposal
* FakeTimers version : 6.0.1
* Environment   : N/A
* Example URL   : N/A
* Other libraries you are using: None

Repo: https://github.com/tc39/proposal-temporal/

We should support faking out usage of all `Temporal.now.*` methods. Not sure if we need to fake anything else?

Note that I do not think we should support `Temporal` _now_ - the proposal is still at stage 2. So consider this more of a tracking issue than something that's actionable right now.

N/A

**What actually happens**
N/A

**How to reproduce**
N/A

SimenB:
Disclaimer: Claude Code has done this work - I just pointed it at docs and suggested APIs 🙂 

## Summary

Implements support for the [Temporal API](https://tc39.es/proposal-temporal/)

Fixes #335

### What's faked

When `"Temporal"` is included in `toFake` (or by default when the host has Temporal), all `Temporal.Now.*` methods are intercepted and resolved against the fake clock:

- `Temporal.Now.instant()`
- `Temporal.Now.timeZoneId()` _(passthrough — fake-timers fakes time, not place)_
- `Temporal.Now.zonedDateTimeISO(tz?)`
- `Temporal.Now.plainDateTimeISO(tz?)`
- `Temporal.Now.plainDateISO(tz?)`
- `Temporal.Now.plainTimeISO(tz?)`

Everything else on `Temporal` (constructors, parsing, arithmetic) is passed through unchanged, so `instanceof Temporal.PlainDate` etc. keep working.

### Temporal.Duration in tick/jump

`clock.tick()`, `clock.tickAsync()`, and `clock.jump()` now accept a `Temporal.Duration` in addition to numbers and `"h:m:s"` strings:

```js
clock.tick(new Temporal.Duration(0, 0, 0, 0, 1, 30)); // 1 hour 30 minutes
clock.jump(Temporal.Duration.from({ days: 1 }));
await clock.tickAsync(Temporal.Duration.from('PT5S'));
```

Calendar units (years, months, weeks, days) are resolved against the clock's current time via `duration.total({ unit: 'millisecond', relativeTo })`.

### Compatibility

Fully guarded behind `isPresent.Temporal` — environments without native Temporal (Node < 26, older browsers) are completely unaffected. Tests skip automatically when Temporal is unavailable.

### Testing

Tests are in the existing `fake-timers-test.js` suite under `describe("Temporal.Now")` and will run on Node 26+. All tests pass on Node 25 (skipped). `npm run lint` passes.


src/fake-timers-src.js
```diff
@@ -541,6 +543,9 @@ function withGlobal(_global) {
     isPresent.clearImmediate =
         _global.clearImmediate && typeof _global.clearImmediate === "function";
     isPresent.Intl = _global.Intl && typeof _global.Intl === "object";
+    isPresent.Temporal =
+        typeof _global.Temporal !== "undefined" &&
+        typeof _global.Temporal.Now !== "undefined";
```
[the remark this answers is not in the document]
fatso83: This is a good point.

fatso83:
Where tf did this failure come from?
```
     TypeError: refute.defined is not a function
      at Context.<anonymous> (test/fake-timers-test.js:6798:23)
```

SimenB:
> Where tf did this failure come from?
> 
> ```
>      TypeError: refute.defined is not a function
>       at Context.<anonymous> (test/fake-timers-test.js:6798:23)
> ```

That is a very good question 😅

SimenB:
@fatso83 ok, I think I'm done now 😀 Unless you spot anything else?

(fun coincidence - I donated blood yesterday, so I thought about when we randomly sat across from one another there during covid 😀)

src/fake-timers-src.js
```diff
@@ -2488,14 +2599,23 @@ function withGlobal(_global) {
...
-         * @param {string|number} tickValue number of milliseconds or a human-readable value like "01:11:15"
+         * @param {string|number|TemporalDuration} tickValue number of milliseconds, a human-readable value like "01:11:15", or a Temporal.Duration
          * @returns {number} the new `now` value
          */
         clock.jump = function jump(tickValue) {
-            const msFloat =
-                typeof tickValue === "number"
-                    ? tickValue
-                    : parseTime(tickValue);
+            let msFloat;
```
fatso83: Is this not _exactly_ the same parsing code as in 2166? Just pull it out into a helper method.

fatso83 (dismissed):
This looks perfect ✨

Just one minor feedback to keep parsing of `number|string|Duration` unified.

src/fake-timers-src.js
SimenB: good catch - pushed 👍

SimenB:
I also added node 26 to CI so we actually run the tests 😅

fatso83:
🇳🇴 True grit! Finally there after 5,5 years 😄

PR #571, opened 2026-05-05
fix: stop `nextAsync` `advanceUntilModeChanges ` from leaking past uninstall

### LINKED ISSUE #562: Race condition bug in nextAsync code?
I have noticed two flaky test runs today. Both only appeared during the Sauce Labs step of one of the workflows and just on the `test-edge` script (but if it fails here, it will short-circuit running subsequent tests)

# First run
[The first was this](https://github.com/sinonjs/fake-timers/actions/runs/24251869152/job/70813136642):
```
  1) FakeTimers
       nextAsync
         stays side-effect free when no timers:

      Error: [assert.equals] 30000000 expected to be equal to 0
      + expected - actual

      -30000000
      +0
      
      at Object.fail (node_modules/@***/referee/lib/create-fail.js:5:20)
      at Object.fail (node_modules/@***/referee/lib/define-assertion/index.js:47:16)
      at assertion (node_modules/@***/referee/lib/define-assertion/index.js:65:10)
      at referee.<computed>.<computed> [as equals] (node_modules/@***/referee/lib/define-assertion/index.js:92:21)
      at Context.<anonymous> (test/fake-timers-test.js:1952:19)

```

Re-running  it made the test pass.

# Second run
[Also in the `nextAsync` bit of the timers](https://github.com/sinonjs/fake-timers/actions/runs/24260861674/job/70844030402):
```
  1) FakeTimers
       setTickMode
         nextAsync
           works with manual calls to async tick functions
             runToLastAsync:

      Error: [assert.equals] [ 1, 2, 3, 4, 5 ] expected to be equal to [ 1, 2, 3, 4 ]
      + expected - actual

         1
         2
         3
         4
      -  5
       ]
      
      at Object.fail (node_modules/@***/referee/lib/create-fail.js:5:20)
      at Object.fail (node_modules/@***/referee/lib/define-assertion/index.js:47:16)
      at assertion (node_modules/@***/referee/lib/define-assertion/index.js:65:10)
      at referee.<computed>.<computed> [as equals] (node_modules/@***/referee/lib/define-assertion/index.js:92:21)
      at Context.<anonymous> (test/fake-timers-test.js:4312:27)
```

These are the only two flaky test runs I can remember for as long as we have been doing this, which makes me a bit wary ...

### LINKED ISSUE #564: Under "nextAsync" tick mode, using tickAsync() without awaiting can cause Mocha process to hang after suite completion
Ran into an interesting issue while trying to upgrade the main sharedb library from `sinon@15` (`@sinonjs/fake-timers@10.3.0`) to the latest `sinon@21` (`@sinonjs/fake-timers@15.3.x`).

I was experimenting with the new `setTickMode({mode: 'nextAsync'})` to resolve post-upgrade test timeouts. While that helped the test cases to no longer time out, I ran into a different issue where the Mocha Node process was left hanging after test suite completion.

A bunch of debugging later, including pointing Gemini at the issue to help me dig into the fake-timers code, I think the cause was using `clock.TickAsync()` without awaiting/then-ing the promise. (I did write this issue and repro without LLM assistance, hope it's not too wordy.)

It _could_ just be user error with using `clock.tickAsync()` without awaiting/then-ing the promise. I probably should've caught that when reviewing those tests initially, but it feels like the clock shouldn't get stuck in an infinite loop after `uninstall()` regardless. 😅 

## Simple repro
_Note that `why-is-node-running` is optional, I included it in case you want to see the open timers yourself._

Save it somewhere, run `npm i --no-save mocha @sinonjs/fake-timers why-is-node-running`, then `npx mocha $FILE`.

```js
const whyIsNodeRunning = require('why-is-node-running').default;
const FakeTimers = require('@sinonjs/fake-timers');

describe('repros for nextAsync + tickAsync hanging Mocha', () => {
  let clock;
  beforeEach(() => {
    clock = FakeTimers.install();
    clock.setTickMode({mode: 'nextAsync'});
  });

  afterEach(() => {
    clock.uninstall();
  });

  after(() => {
    setImmediate(() => whyIsNodeRunning());
  });

  // This causes the Mocha process to hang after test suite completion
  it.only('tickAsync = setTimeout', (done) => {
    setTimeout(() => done(), 500);
    clock.tickAsync(500);
  });
});
```

The issue also manifests when tickAsync is greater than the timeout. It does not manifest when tickAsync < timeout, nor does it manifest when using the synchronous `tick()`.

I ran this repro under Node v24.11.1, but I also saw hangs with Node 20 and 22 in the sharedb CI runs.

## Probable explanation

_Code permalinks here point to the latest commit as of 15.13.1, published earlier today._

1. `tickAsync()` temporarily sets the tick mode to "manual" in `pauseAutoTickUntilFinished()`, then runs waiting timers.
    - https://github.com/sinonjs/fake-timers/blob/7cda826a9ed8413f72e8df87088c655a4b4f1f0d/src/fake-timers-src.js#L1854-L1862
2. A timer execution causes the test case to finish immediately, e.g. via calling `done()`.
3. `afterEach` runs `clock.uninstall()`.
4. The test case completes, and Mocha prints the suite summary since there are no more test cases. Normally, the Node process would automatically exit at this point.
5. However, now inside `pauseAutoTickUntilFinished()` above, the `promise.finally()` runs, which calls `setTickMode("nextAsync")` on the uninstalled clock.
6. That final `setTickMode("nextAsync")` calls `advanceUntilModeChanges()`, and the end of that function enters into the infinite async loop below, since the clock is uninstalled and no longer changes `tickMode.counter`.
    - https://github.com/sinonjs/fake-timers/blob/7cda826a9ed8413f72e8df87088c655a4b4f1f0d/src/fake-timers-src.js#L1839-L1845

I verified in a debugger that Node in fact gets infinitely stuck in that loop.

## Fixes?

For my case, switching fully to `clock.tick()` resolves the issue, but like I mentioned above, fake-timers could potentially be more defensive to prevent the infinite loop, especially for cases that legitimately need `tickAsync()`.

The Gemini-suggested fix was to set a `clock.uninstalled = true` flag on `uninstall()`, and in `pauseAutoTickUntilFinished()`, only run `setTickMode("nextAsync")` if the clock is not uninstalled. That did resolve the issue for me, both in the isolated repro and in the full sharedb test suite.

However, I'm wondering... would it make more sense to update `setTickMode()` to check if the clock has been uninstalled, and early return? Or maybe throw an error in that case? (That would also require the `pauseAutoTickUntilFinished()` guard above.)

I'd be willing to make a PR if needed, since I'm doing a bunch of open-source work right now anyways.

SimenB:
@fatso83 hey again! I chucked Claude at #562 and it believes it found it, plus the #564 error. Here's its summary:

## Summary

Fixes the process-hang bug (#564) and the flaky `nextAsync` test failures (#562).

### Bug 1 — infinite loop after `uninstall()` (#564)

`pauseAutoTickUntilFinished` saves the current tick mode and restores it in `.finally()` when the async operation settles. The problem: if `clock.uninstall()` runs before `.finally()` fires (which happens whenever a timer callback calls `done()` or resolves a test promise — causing the test to finish and `afterEach` to uninstall the clock while the un-awaited `tickAsync` is still in flight), the `.finally()` calls `setTickMode({ mode: "nextAsync" })` on the now-dead clock. This starts a new `advanceUntilModeChanges` loop whose counter can never change, spinning forever and keeping the process alive after the suite exits.

**Fix:** move `clock.uninstall` into `createClock` so it can close over a private `let uninstalled = false` flag. Setting `uninstalled = true` at the top of `clock.uninstall` is enough: `pauseAutoTickUntilFinished`'s `.finally()` checks `!uninstalled` before restoring `nextAsync` mode.

### Bug 2 — flaky `runToLastAsync` assertion (#562)

Mocha inserts macrotask boundaries between `beforeEach` hooks and before running the test body. The outer `beforeEach` calls `setTickMode({ mode: "nextAsync" })`, starting `advanceUntilModeChanges` (AUMC). Each time Mocha yields to schedule the next hook or the test body, AUMC can complete another iteration and call `clock.next()`. If AUMC fires timer 4 (at T=5, which dynamically schedules timer 5 at T=6) *before* the test body reaches `runToLastAsync()`, then `lastTimer` returns T=6 — and `runToLastAsync` includes timer 5 in its run, producing `[1,2,3,4,5]` instead of the expected `[1,2,3,4]`.

**Fix:** in the inner `beforeEach`, reset to `"manual"` mode before adding timers so setup is in a known state. Then re-enable `"nextAsync"` as the first synchronous line of each test body — before any `await` — so AUMC has not had a chance to advance the clock when the async method is called.


SimenB:
Hmm, copilot's comment is a good spot. I wonder if we can move things around to avoid the new property

src/fake-timers-src.js
```diff
@@ -2639,6 +2563,97 @@ function withGlobal(_global) {
             clock.hrtime = hrtime;
         }
 
+        /**
+         * @returns {Timer[]}
+         */
+        clock.uninstall = function () {
```
SimenB, line 2577: implementation is the same - just moved to close over `uninstalled`

src/fake-timers-src.js
```diff
@@ -1996,7 +1918,9 @@ function withGlobal(_global) {
             }
             clock.setTickMode({ mode: "manual" });
             return promise.finally(() => {
-                clock.setTickMode({ mode: "nextAsync" });
+                if (!uninstalled) {
```
SimenB, line 1926: this is the fix

src/fake-timers-src.js
```diff
@@ -2639,6 +2563,97 @@ function withGlobal(_global) {
             clock.hrtime = hrtime;
         }
 
+        /**
+         * @returns {Timer[]}
+         */
+        clock.uninstall = function () {
+            uninstalled = true;
+            clock.setTickMode({ mode: "manual" });
```
SimenB, line 2579: these two lines are new - otherwise the same as before

fatso83:
Might want to do a rebase to catch any type regressions 😛

SimenB:
merged in main. still passing 🥳

fatso83 (approved):

fatso83:
Two biggos in one evening. Nais.

fatso83:
I think a test broke that this was supposed to fix?

```
  1) FakeTimers
       setTickMode
         nextAsync
           works with manual calls to async tick functions
             runToLastAsync:

      Error: [assert.equals] [ 1, 2, 3, 4, 5 ] expected to be equal to [ 1, 2, 3, 4 ]
      + expected - actual

         1
         2
         3
         4
      -  5
       ]
```

https://github.com/sinonjs/fake-timers/actions/runs/25405330189/job/74514771134

PR #573, opened 2026-05-05
ci: add types job to workflow

fatso83:
Ensure types:check and types:smoke are run during CI. Type validation
was previously only in posttest, which was bypassed by individual CI jobs.

This made us miss that the Temporal PR (#569) had breaking types 😑

Having a dedicated step should avoid that in the future.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,6 +64,26 @@ jobs:
               run: |
                   npm run lint
 
+    types:
+        runs-on: ubuntu-latest
+        steps:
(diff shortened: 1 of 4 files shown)


SimenB:
hide

fatso83:
LOL, as if I would not do the same. With no checks to warn you, how are you supposed to know? Gemini got it covered, though 👍

SimenB:
Both Claude and Copilot missed it 😀 Nice to have 3 bots, haha

PR #574, opened 2026-05-05
fix: accept duck-typed `TemporalDuration` when `Temporal` absent from global

SimenB:
I tried to use the new `Temporal` stuff in Jest, and hit this 😀

`tickValueToMs` gated the `Duration` duck-type check behind `isPresent.Temporal`, so any object with a `.total()` method fell through to `parseTime()` and threw `TypeError: str.split is not a function` when Temporal was absent from the faked global.

The guard doesn't belong here. Accepting a Duration as input to `tick()` is purely duck-typed - `durationToMs` just calls `duration.total(...)`, it has no dependency on the native Temporal global. The `isPresent.Temporal` guard is correct for faking `Temporal.Now.*` (which has to construct real Temporal objects as return values), not for accepting input.

Two changes:
- Remove `isPresent.Temporal &&` from the duck-type check in `tickValueToMs`
- Make `durationToMs` omit `relativeTo` when `NativeTemporal` is unavailable — calendar-unit durations will throw from within `.total()` itself, which is the right place for that error

`tick()`, `tickAsync()`, and `jump()` all go through `tickValueToMs` so all three get the fix.

### DIFF
--- a/src/fake-timers-src.js
+++ b/src/fake-timers-src.js
@@ -2085,10 +2085,18 @@ function withGlobal(_global) {
         function durationToMs(duration) {
             // relativeTo uses the real system timezone — fake-timers fakes time, not place.
             // Calendar-unit durations (months, years) will resolve DST/length using the host tz.
-            const relativeTo = NativeTemporal.Instant.fromEpochMilliseconds(
-                clock.now,
-            ).toZonedDateTimeISO(NativeTemporal.Now.timeZoneId());
-            return duration.total({ unit: "millisecond", relativeTo });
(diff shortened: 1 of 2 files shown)


fatso83 (approved):

fatso83:
agh, clumsy tired fingers. I was too trigger happy. I saw this might have pending changes. Just re-open and issue a revert on my merge if you need to (if you wanted to address something else)

This is my sign to tuck in. God natt 💤

SimenB:
I'll take a look tomorrow. Natta 😀

PR #575, opened 2026-05-05
test: fix `runToLastAsync` flake by avoiding AUMC race in `nextAsync` mode

### LINKED ISSUE #562: Race condition bug in nextAsync code?

SimenB:
First commit tries to fix https://github.com/sinonjs/fake-timers/pull/571#issuecomment-4383589793

Second addresses https://github.com/sinonjs/fake-timers/pull/574#discussion_r3191891022

---

~I'll dig some more into the test failure tho (but tomorrow)~

Closes #562. Completes the partial fix in #571.

## Diagnosis

`runToLastAsync` is documented as a bounded operation: "advances the clock to that time firing callbacks as necessary. If new timers are added while it is executing they will be run only if they would occur before this time" (README). The implementation routes through `pauseAutoTickUntilFinished`, which restores `nextAsync` mode in `.finally()`. After the operation resolves, AUMC restarts and fires the dynamically-queued timer the docs say should not run — directly contradicting the contract. In Edge, AUMC's macrotask wins the race against the post-await assertion, making the violation observable as the SauceLabs flake.

#571 addressed the _before_ race (AUMC advancing during Mocha's hook→test macrotask boundary) with a test-only mitigation. It did not address the _after_ race introduced by the auto-restore in `.finally()`. The flake reappeared on Edge after #571 merged for exactly this reason.

## Fix

`runToLastAsync` now opts out of the auto-restore: it pauses AUMC for the duration of the call and leaves the clock in `manual` mode on completion. Callers re-enable `nextAsync` explicitly if they want auto-advancement to resume. This is the only call site where the documented bounded semantic conflicts with AUMC restarting — `tickAsync`/`nextAsync`/`runAllAsync` keep the existing restore behaviour.

**Implementation:** adds a `restoreNextAsync` parameter to `pauseAutoTickUntilFinished` / `runAsyncWithNativeTimeout`; only `runToLastAsync` passes `false`. The `uninstalled` guard from #571 is preserved on the restoring path.

## Behaviour change

Users in `nextAsync` mode who relied on AUMC continuing after `runToLastAsync` need to add `clock.setTickMode({ mode: "nextAsync" })` after the await. README updated with a note. Minor version bump.

### DIFF
--- a/test/fake-timers-test.js
+++ b/test/fake-timers-test.js
@@ -4212,8 +4212,8 @@ describe("FakeTimers", function () {
 
             it("does not create an infinite loop when uninstalled while a tickAsync is in flight", function (done) {
                 // Regression test for https://github.com/sinonjs/fake-timers/issues/564.
-                // tickAsync() pauses nextAsync mode; the restore is scheduled via a
-                // native setTimeout. If uninstall() runs before that fires, the restore
+                // tickAsync() pauses nextAsync mode; the restore runs via a native
+                // macrotask (setImmediate/setTimeout). If uninstall() runs before that fires, the restore
                 // must be a no-op.
                 setTimeout(done, 100);
                 clock.tickAsync(100); // intentionally NOT awaited
@@ -4311,6 +4311,7 @@ describe("FakeTimers", function () {
                 });
 
                 afterEach(async function () {
+                    clock.setTickMode({ mode: "nextAsync" });
                     await allTimersDone;
                     assert.equals(timerLog, [1, 2, 3, 4, 5]);
                 });
@@ -4328,8 +4329,7 @@ describe("FakeTimers", function () {
                     assert.equals(timerLog, [1, 2, 3, 4]);
                     // mode is left manual so AUMC cannot fire timer 5
                     assert.equals(clock.tickMode.mode, "manual");
-                    // re-enable so afterEach can drain the remaining timers
-                    clock.setTickMode({ mode: "nextAsync" });
+                    // afterEach re-enables nextAsync to drain remaining timers
                 });
 
                 it("nextAsync", async function () {
--- a/test/fake-timers-test.js
+++ b/test/fake-timers-test.js
@@ -4323,12 +4323,12 @@ describe("FakeTimers", function () {
                 });
 
                 it("runToLastAsync", async function () {
-                    clock.setTickMode({ mode: "nextAsync" });
+                    // Do not enable nextAsync before this call: pauseAutoTickUntilFinished
+                    // restores it in .finally(), and in some engines AUMC can fire
+                    // clock.next() before the post-await assertion runs.
                     await clock.runToLastAsync();
                     // 5 should not resolve because it wasn't queued when we called "only pending timers"
                     assert.equals(timerLog, [1, 2, 3, 4]);
-                    // mode is left manual so AUMC cannot fire timer 5
-                    assert.equals(clock.tickMode.mode, "manual");
                     // afterEach re-enables nextAsync to drain remaining timers
                 });
 
--- a/test/fake-timers-test.js
+++ b/test/fake-timers-test.js
@@ -6642,8 +6642,7 @@ describe("missing timers", function () {
         });
 
         it("tick() accepts a TemporalDuration-shaped object even when Temporal is absent from the global", function () {
-            // Use a custom global without Temporal so isPresent.Temporal is false,
-            // ensuring the test exercises the fix even on Node 26+ where Temporal exists.
+            // Use a custom global without Temporal so isPresent.Temporal is false
             const clock = FakeTimers.withGlobal({
                 Date: globalThis.Date,
             }).createClock(0);
--- a/test/fake-timers-test.js
+++ b/test/fake-timers-test.js
@@ -4326,17 +4326,9 @@ describe("FakeTimers", function () {
                     await clock.runToLastAsync();
                     // 5 should not resolve because it wasn't queued when we called "only pending timers"
                     assert.equals(timerLog, [1, 2, 3, 4]);
-                    // runToLastAsync leaves the clock in manual mode; re-enable
-                    // nextAsync so afterEach can drain the remaining timers.
-                    clock.setTickMode({ mode: "nextAsync" });
-                });
-
-                it("runToLastAsync leaves clock in manual mode", async function () {
-                    clock.setTickMode({ mode: "nextAsync" });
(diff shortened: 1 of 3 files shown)


atscott:
@SimenB I feel like the approach doesn't match up with my expectations for how things would or should work. 

My interpretation of the documentation is that the contract of `runToLastAsync` is to run timers up to the bound during its execution. What happens after it completes is the responsibility of the active mode (`nextAsync`). The flake happens because the auto-tick (macrotask) races with the test's continuation (microtask) after the await. A less invasive fix would be to update the test to handle this race (e.g., by asserting `clock.now` instead of checking the log length immediately, or adjusting the assertions) rather than changing the behavior of `runToLastAsync`. 

IIUC, there's no real race condition for the code immediately following the `await runToLastAsync` and the bounded guarantees are correct. The code following `await runToLastAsync` resumes in a `microtask`. The auto-ticking in `nextAsync` mode schedules its next step via a `macrotask` so code following the `runToLastAsync` should reliably run before the clock advances.

This also feels like it applies equally to all the other "async" tick methods that could (but don't) have similar documentation about bounded guarantees, and they would suffer from the exact same potential race conditions if they restore nextAsync mode on completion. So one could also argue this is a documentation problem.

If this change is still deemed correct, I would feel that it should consider disabling the "interval" tick as well. If interval is set to 1 and the execution of `runToLastAsync` takes longer than 1ms, I would think there would be a similar race condition with the automatic advancement there, right? 

The "temporarily disable `nextAsync`" was intended to prevent double ticks _during_ the executions of the async ticks, but not intended to have a side effect of changing the tick mode after they're called.

SimenB:
Fair enough 🙂 Pushed a test only change as I honestly don't understand the feature and probably won't expose it in Jest due to that 😅 But if the tests now assert the behaviour you expect without flake that's an improvement 👍

SimenB:
@fatso83 PTAL when you have time - I'm only waiting for release of #574 before making a release of Jest 😀

atscott:
I definitely don't mean to shut down the conversation. Honestly ticking manually when `nextAsync` is enabled is a bit weird on its own and there could be many approaches (throwing, always interpreting is a request to switch to "manual", etc).

I think I'm mostly confused by the comment in the test "Do not enable nextAsync before this call: pauseAutoTickUntilFinished restores it in .finally(), and in some engines AUMC can fire clock.next() before the post-await assertion runs." and I guess the related failure in #562. It seems to indicate that somehow the resuming of `nextAsync` in the finally still does result in another `clock.next` before the test assertions after the `await runToLastAsync` and I don't understand how that's possible.

atscott:
Do you know if the saucelabs test running the legacy edge or is it the chromium based edge? Legacy edge did have issues with setImmediate and microtasks (https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/).

SimenB:
Hmm, good question! I don't have access there, so we'd have to wait for @fatso83 to check I think

fatso83:
Sorry ... forgot this hanging on me. I verified the [config](https://github.com/sinonjs/fake-timers/blob/main/mochify.webdriver.js) on https://saucelabs.com/platform-configurator, and MicrosoftEdge is the name used to get the current Edge browser based on Chromium.

test/fake-timers-test.js
```diff
@@ -4321,10 +4323,13 @@ describe("FakeTimers", function () {
                 });
 
                 it("runToLastAsync", async function () {
-                    clock.setTickMode({ mode: "nextAsync" });
+                    // Do not enable nextAsync before this call: pauseAutoTickUntilFinished
```
atscott, line 4326: I might suggest deleting the test entirely. Omitting the `clock.setTickMode({ mode: "nextAsync" });` makes the test not really cover anything that's not already covered elsewhere. 

An alternative would be to verify the background mode-transition rather than the result of the timers (a little big more implementation-detail testing but I really can't think of anything else):

```ts
it("runToLastAsync pauses and restores nextAsync mode", async function () {
    clock.setTickMode({ mode: "nextAsync" });
    const spy = sinon.spy(clock, "setTickMode");
    // Only queue one timer so there are no subsequent timers to trigger AUMC races after completion
    setTimeout(() => timerLog.push(1), 1);
    await clock.runToLastAsync();
    // Verify it ran the timer
    assert.equals(timerLog, [1]);
    // Verify it paused (set to manual) and restored (set to nextAsync)
    assert.isTrue(spy.calledWith({ mode: "manual" }));
    assert.isTrue(spy.calledWith({ mode: "nextAsync" }));
    spy.restore();
});
```

